PR #12, opened 2024-02-12
fix: import dayjs plugins from esm not directly

### LINKED ISSUE #1: Fix build (dayjs import) for tests and production
If we run

```bash
npm run build
npm run test
```

we get the error

```
TypeError: dayjs.extend is not a function
```

Similarly, we get an error if we build the storybook with

```bash
npm run build
npm run build-storybook
```

[Opening the prevalence over time plot](https://genspectrum.github.io/dashboards/?path=/story/components-prevalence-over-time--monthly), we get:

image

However, **the local development instance of storybook works:**

```bash
npm run build
npm run storybook
```

JonasKellerer:
Resolves #1 

The solution comes from https://github.com/iamkun/dayjs/issues/1242

### DIFF
--- a/components/src/temporal.ts
+++ b/components/src/temporal.ts
@@ -1,6 +1,7 @@
-import * as dayjs from 'dayjs';
-import * as isoWeek from 'dayjs/plugin/isoWeek';
-import * as advancedFormat from 'dayjs/plugin/advancedFormat';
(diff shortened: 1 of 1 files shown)


fengelniederhammer (approved):
Seems to work :+1:

PR #14, opened 2024-02-12
ci: add workflow for linting, formatting and testing

### LINKED ISSUE #2: Run tests in CI
We should run the tests in GitHub actions after fixing #1.

JonasKellerer:
Resolves #2

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,51 @@
+name: Tests
+
+on: [push]
+
+defaults:
+  run:
+    working-directory: ./components
+
+jobs:
+  checks:
+    name: Run Checks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+
+      - uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('components/**/package-lock.json') }}
+
+      - name: Install npm packages
+        run: npm ci
+
+      - name: Get Installed Playwright Version
+        id: playwright-version
+        run: echo "PLAYWRIGHT_VERSION=$(node -e "console.log(require('components/package-lock.json').packages['node_modules/@playwright/test'].version)")" >> $GITHUB_ENV
+
+      - name: Cache Playwright Browsers
+        uses: actions/cache@v4
+        id: playwright-cache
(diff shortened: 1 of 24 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,51 @@
...
+
+on: [push]
+
+defaults:
+  run:
+    working-directory: ./components
+
+jobs:
+  checks:
+    name: Check format
```
fengelniederhammer, line 11:
```suggestion
    name: Run tests
```

components/src/types.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type LapisFilter = Record<string, any>;
```
fengelniederhammer, line 2:
```suggestion
export type LapisFilter = Record<string, unknown>;
```

components/src/operator/DivisionOperator.ts
```diff
@@ -9,7 +9,7 @@ export class DivisionOperator<
     R extends string,
     N extends string,
     D extends string,
-> implements Operator<{ [P in K]: S[K] } & MappedNumber<R | N | D>>
+> implements Operator<{ [P in K]: S[K] } & MappedNumber<R> & MappedNumber<N> & MappedNumber<D>>
```
fengelniederhammer, line 12: This might be worth to extract it:
```
type DivisionOperatorResult<K extends keyof S, S, R extends string, N extends string, D extends string> = {
    [P in K]: S[K];
} & MappedNumber<R> &
    MappedNumber<N> &
    MappedNumber<D>;
```

components/src/operator/DivisionOperator.ts
```diff
@@ -24,7 +24,7 @@ export class DivisionOperator<
     async evaluate(
         lapis: string,
         signal?: AbortSignal,
-    ): Promise<Dataset<{ [P in K]: S[K] } & MappedNumber<R | N | D>>> {
+    ): Promise<Dataset<{ [P in K]: S[K] } & { [P in R]: number } & { [P in N]: number } & { [P in D]: number }>> {
```
fengelniederhammer, line 27: Can't we also simply write this as `DivisionOperatorResult`? `{ [P in R]: number }` should be equal to `MappedNumber<R>`?

.github/workflows/tests.yml
fengelniederhammer, line 1: - [ ] Reminder: squash

PR #46, opened 2024-03-09
refactor: use daisyui for tabs and containers

JonasKellerer:
### Summary
- refactor container components to make them more modular
- more use of daisyui features like tabs and dropdown

### Screenshot
grafik

### PR Checklist
(There are no tests yet)


components/src/components/container/component-checkbox-selector.ts
```diff
@@ -0,0 +1,75 @@
...
+                <div tabindex="0" role="button" class="btn btn-${this.size}">${this.label}</div>
+                <ul tabindex="0" class="p-2 shadow menu dropdown-content z-[1] bg-base-100 rounded-box w-52">
+                    ${this.checkedItems.map(
+                        (item, index) => html`
+                            <li>
+                                <label>
+                                    <input
+                                        type="checkbox"
+                                        ?checked=${item.checked}
+                                        @change=${() => {
```
JonasKellerer, line 56: I'm not sure if this is a nice design. What do you think @fengelniederhammer ?

The idea was, that this element can be initialized through its props, but changes are handled internally and only the change is broadcasted through the Event. 
However, we need to have the same structure of state and setState in the parent component, which uses this component (see `mutations.ts`). So it might be easier to use this component without the internal state.

chaoran-chen:
Would it be possible to leave the tabs at the bottom while minimize the whitespace

chaoran-chen:
Sorry for the very brief comment earlier. The tabs look much better than the previous ones!

To explain what I would like to improve, something that I don't like about the CoV-Spectrum plots is that the actual plots are really small: in this example, the plot only covers 36.8% of the area of a component:

image

With the new design, I tried to minimize the whitespace. Even with the very large title, the plot covers 71% of the area. Hereby, it was a conscious decision to have the tabs on the same row as the scale options (linear/logarithmic) and the question mark button.

image

With the design of this PR, there is more whitespace and the plot only covers 57.3% of the area and I would like to increase the number.

image

components/src/components/container/component-checkbox-selector.ts
```diff
@@ -0,0 +1,75 @@
...
+                };
+            }
+            return checkedItem;
+        });
+    }
+
+    override render() {
+        return html`
+            <div class="dropdown">
+                <div tabindex="0" role="button" class="btn btn-${this.size}">${this.label}</div>
```
fengelniederhammer, line 47:
```suggestion
                <div tabindex="0" role="button" class="btn ${this.size}">${this.label}</div>
```
I quite sure that optimization for only actually used css classes doesn't work, if we use them dynamically. Is the `size` prop actually used currently? We can also remove it for now and check later what customization abilities we need here.

components/src/components/container/component-checkbox-selector.stories.ts
fengelniederhammer, line 1: Do we want to add a basic test for this component?

components/src/components/container/component-container.ts
```diff
@@ -1,149 +1,18 @@
...
-                    @toolbarclick="${(event: CustomEvent) => {
-                        if (event.detail === 'info') {
-                            this.toggleInfo();
-                        }
-                    }}"
-                >
-                    <slot name="toolbar"></slot>
-                </div>
+            <div>
+                <p class="text-2xl m-4">${this.heading}</p>
```
fengelniederhammer, line 13: Let's use a `h1` or `h2` tag instead?

components/src/components/container/component-container.ts
```diff
@@ -1,149 +1,18 @@
-import { LitElement, html, css } from 'lit';
-import { customElement, property, queryAssignedElements } from 'lit/decorators.js';
-import { ComponentTab } from './component-tab';
-import { ComponentToolbar } from './component-toolbar';
+import { html } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { TailwindElement } from '../../tailwind-element';
 
 @customElement('gs-component-container')
```
fengelniederhammer, line 5: The name doesn't really fit anymore, but we can also change that later.

components/src/components/container/component-info.stories.ts
```diff
@@ -0,0 +1,18 @@
...
+    component: 'gs-component-info',
+};
+
+export default meta;
+
+export const InfoStory: StoryObj = {
+    render: () => {
+        const content = 'This is a tooltip which shows some information.';
+
+        return html` <gs-component-info content=${content}></gs-component-info> `;
```
fengelniederhammer, line 16: grafik

The tooltip isn't visible (at least in my Firefox)

components/src/components/container/component-tabs.ts
```diff
@@ -0,0 +1,36 @@
...
+
+        const tabElements = this.tabs.map((tab, index) => {
+            return html`
+                <input
+                    type="radio"
+                    name=${tabNames}
+                    role="tab"
+                    class="tab"
+                    aria-label=${tab.title}
+                    ?checked=${index === 0}
```
fengelniederhammer, line 26: What does the `?` do here?

components/src/components/container/component-info.ts
```diff
@@ -1,22 +1,15 @@
...
-
-    @property({ type: Boolean, reflect: true })
-    active = false;
+export class ComponentInfo extends TailwindElement() {
+    @property({ type: String })
+    content: string = '';
 
     override render() {
-        return html`<slot></slot>`;
+        return html`<div class="tooltip" data-tip=${this.content}>
```
fengelniederhammer, line 11: I can imagine that a tooltip isn't the right tool to display those info texts. Some are probably larger that a usual tooltip. But let's see when we write actual info texts.

components/src/components/mutations.stories.ts
```diff
@@ -27,11 +27,13 @@ export default meta;
 const Template: StoryObj<MutationsProps> = {
     render: (args) => html`
         <gs-app lapis="${LAPIS_URL}">
-            <gs-mutations
-                .variant=${args.variant}
-                .sequenceType=${args.sequenceType}
-                .views=${args.views}
-            ></gs-mutations>
+            <div class="max-w-screen-sm">
```
fengelniederhammer, line 30: We should probably instead make sure that the tab itself isn't larger than necessary.

components/src/components/mutations.ts
```diff
@@ -24,15 +23,13 @@ export type MutationsProps = {
...
+};
+
 @customElement('gs-mutations')
 export class Mutations extends LitElement {
-    static override styles = css`
-        .content {
-            max-height: 300px;
-            overflow-y: auto;
-        }
-    `;
```
fengelniederhammer, line 34: At some point, we should bring this back in some form (maybe as a separate component?)

components/src/query/queryRelativeGrowthAdvantage.ts
```diff
@@ -3,6 +3,43 @@ import { FetchAggregatedOperator } from '../operator/FetchAggregatedOperator';
 import { MapOperator } from '../operator/MapOperator';
 import { getMinMaxTemporal, TemporalCache, YearMonthDay } from '../temporal';
 
+export type RelativeGrowthAdvantageData = {
```
fengelniederhammer, line 6: We can probably also use this type somewhere in this file? Either `export type RelativeGrowthAdvantageData = ReturnType<typeof queryRelativeGrowthAdvantage>` or declaring the response type explicitly, if TS can't infer the response type properly.

components/src/components/relative-growth-advantage.ts
```diff
@@ -45,49 +44,61 @@ export class RelativeGrowthAdvantage extends LitElement {
...
+            pending: () => html` <gs-component-container heading=${this.heading}> Loading...</gs-component-container> `,
             complete: (data) => {
                 if (data === null) {
-                    return html`
-                        <h1>Relative growth advantage</h1>
-                        <p>No data available.</p>
-                    `;
+                    return html` <gs-component-container heading=${this.heading}>
+                        >No data available.</gs-component-container
+                    >`;
```
fengelniederhammer, line 89:
```suggestion
                    return html` <gs-component-container heading=${this.heading}>
                        No data available.</gs-component-container
                    >`;
```
And the formatting looks quite broken ;)

fengelniederhammer (changes requested):
Looks good in general :)

* Do we want to write some basic tests for the helper components?
* How do we handle that issue that subcomponents create their own shadow DOM? In this PR or in a separate issue?

components/src/components/container/component-checkbox-selector.ts
JonasKellerer: It is not used. I added it, because we might want to change the size of the text somehow later on. We have many different sizes right now. I removed it here and we can add it later on.

components/src/components/container/component-checkbox-selector.stories.ts
JonasKellerer: We do, but there were no tests available at that time ;) I'll add some.

components/src/components/container/component-container.ts
JonasKellerer: I changed it to h2. h1 seems to me for the heading of the page. The margin is now also gone. However we might want to add this later on with tailwind/typography https://daisyui.com/docs/layout-and-typography/

chaoran-chen: Hmm, do you know what is the right way to do this in web components? In a normal document, `h1` should be the main title, but I don't know how things should be for components that could be placed at various places (within a sub-subsection that might have a `h4` header or directly as the only content of a page where a `h1` would make sense).

components/src/components/container/component-info.stories.ts
JonasKellerer: For me its the same. However, the tooltip should open to the top. Daisyui does not render the tooltip depending on the given space to other directions. That is why it cant be seen in the story. I will change the story, so one can see the content. However, I would argue to leave the component it like it is since you suggested to change the info object later on.

components/src/components/container/component-info.ts
JonasKellerer: I also think we will need other components to show this. This can only show text. But maybe we want to render also some components.

components/src/components/mutations.ts
JonasKellerer: I think a limiting component would be nice. I added a ticket #55

components/src/query/queryRelativeGrowthAdvantage.ts
JonasKellerer: I used the ReturnType. However, I could not use it inside, since the type inside has for estimatedAbsoluteNumbers.t type number[] and not type YearMonthDay. But this problem is deeper inside the code. My personal feeling is, that we could use always the unix time inside the code, and only have the user facing side using the temporal objects. But this would be a larger refactor. 
Additionally, I think we should use zodios for parsing the fetch results. However, this would be its own ticket.

components/src/components/container/component-container.ts
JonasKellerer: Good question. Currently, I can imagine two scenarios:
1: 
  <ExternalComponent>
    <OurComponent />
    <SecondExternal />
   </ExternalComponent>
 2:
   <ExternalComponent>
    <OurComponent >
      <SecondExternal>
    </OurComponent>
   </ExternalComponent>
   
 In case 1 we can set the shadow root to OurComponent. Then the css we use stays inside. But we could also think about passing css to our component. (Or this is what I've read so far). This way users can style our components as they wish to.
 In case 2 I think we have more of a problem, since the SecondExternal component can have also a `h2` component. We can try to bypass this if we use the slot system. In this case the global css will be used for the SecondExternal component (if I read https://css-tricks.com/styling-a-web-component/ correctly).
 But maybe we want to use both versions.

components/src/components/container/component-checkbox-selector.ts
JonasKellerer: As discussed we pass the set method to the component, so we dont have events bubbling through the tree.

fengelniederhammer (approved):

Issue #60, opened 2024-03-13
Mutations: Add a download button


fengelniederhammer:
Add a download button to download the data as a TSV (same data as shown in the table)

PR #71, opened 2024-03-14
feat(components): use gridjs for mutation table

### LINKED ISSUE #16: Table: Add basic styling

### LINKED ISSUE #17: Table: Allow users to sort by any of the columns

JonasKellerer:
resolves #17, resolves #16

### Summary

We now use [grid.js](https://gridjs.io/) to display tables. I comes out of the box with sorting, formatting, (and search)
This replaces the tables in:
- mutations-grid.ts
- mutations-table.ts
- prevalence-over-time-table.ts

### Screenshot

grafik

### PR Checklist


components/src/components/mutations/mutations-grid.ts
```diff
@@ -1,136 +1,161 @@
...
-                        `;
-                    })}
-                </tbody>
-            </table>
-        `;
+        return orderedPositionsToProportionAtBase.map((proportionsForBaseAtPosition) => {
+            return [
+                this.sequenceType === 'amino acid'
+                    ? proportionsForBaseAtPosition.position
+                    : Number(proportionsForBaseAtPosition.position),
```
fengelniederhammer, line 139: We will have to revisit this in the multi-segment case, don't we?

components/src/components/mutations/mutations-grid.ts
```diff
@@ -1,136 +1,161 @@
...
+                sort: {
+                    compare: (a: string | number, b: string | number) => {
+                        return this.sortPositionsWithSegments(a.toString(), b.toString());
+                    },
+                },
+            },
+            ...bases[this.sequenceType].map((base) => {
+                return {
+                    name: base,
+                    sort: true,
```
fengelniederhammer, line 37: I can click on those columns to "sort" them, but nothing happens...

components/src/components/prevalenceOverTime/prevalence-over-time-table.ts
```diff
@@ -13,29 +14,41 @@ export class PrevalenceOverTimeTable extends TailwindElement() {
...
+        return [
+            {
+                name: this.granularity,
+                sort: true,
+            },
+            ...this.getSplitHeaders(),
+        ];
+    }
+
+    getSplitHeaders() {
```
fengelniederhammer, line 27:
```suggestion
    private getSplitHeaders() {
```

components/src/components/prevalenceOverTime/prevalence-over-time-table.ts
```diff
@@ -13,29 +14,41 @@ export class PrevalenceOverTimeTable extends TailwindElement() {
...
+                    },
+                ],
+            };
+        });
+    }
 
-        const keys = Object.keys(tableData[0]);
+    getData() {
+        const dataByHeader = getPrevalenceOverTimeTableData(this.data, this.granularity);
+        return Object.values(dataByHeader).map((row) => Object.values(row));
```
fengelniederhammer, line 47: Nice, that was easier than I expected :+1:

components/src/components/container/component-table.ts
fengelniederhammer, line 1: grafik

Nothing we need to take care of now, but to keep it in mind: The pagination doesn't look good in dark mode. I can see the numbers when I hover them though.

components/src/components/mutations/mutations-grid.ts
```diff
@@ -1,136 +1,161 @@
...
-    @property({ type: Object })
-    data: Dataset<MutationEntry> | null = null;
-
-    @property({ type: String })
-    sequenceType: SequenceType = 'nucleotide';
+        return {};
+    }
 
-    override render() {
+    getTableData(): (string | number | MutationCell)[][] {
```
fengelniederhammer, line 89: That might also be worth extracting and giving it a unit test (not necessarily now)

fengelniederhammer (approved):
Nice. That looks astonishingly simple :+1:

components/src/components/mutations/mutations-grid.ts
JonasKellerer: I think so. We can also make use of the same sort  as for the amino acid case and remove that here. What do you think?

fengelniederhammer: Sounds good.

components/src/components/mutations/mutations-grid.ts
JonasKellerer: I think this will come with the download button #60

components/src/components/prevalenceOverTime/prevalence-over-time-table.ts
JonasKellerer: Same :)

components/src/components/container/component-table.ts
JonasKellerer: Should we add a ticket for the dark theme issues?

fengelniederhammer: Probably makes sense, yes. Maybe we can simply disable it for now.

components/src/components/mutations/mutations-grid.ts
JonasKellerer: I added a workaround, since gridjs can not cope with a formatter and sorting of object cells. It seems to be a bug.

components/src/components/container/component-table.ts
JonasKellerer: I disabled it.

Issue #81, opened 2024-03-20
Date range selector: Dispatch event


JonasKellerer:
This is a follow up to #9

PR #82, opened 2024-03-20
feat(components): add date range selector and unify selector component

### LINKED ISSUE #9: Overview: Date range filter
We could replicate the implementation on CoV-Spectrum:

Important capabilities are:

- Select special data ranges such as "2023" or "past 2 weeks"
- Select a date through a date picker.
- Type in a date in the ISO format (i.e., `yyyy-mm-dd`)
- Edit a date in the date field directly (e.g., change `2023-09-17` to `2023-10-17` by selecting `09` in the field and then typing `10`)
- Copy&paste a date in the ISO format into a date field

## Tasks

tbd

JonasKellerer:
Resolves #9

### Summary
This adds a date range selector element. In this element you can pick a preset date range, or pick dates by hand.
The date picker dates are limited to the max/min value of the from/to picker.

There is a follow up ticket #81 to dispatch an event, when the date is changed. This should then also introduce a submit button (similar to the location filter)

### Screenshot
image

### PR Checklist


components/src/components/dateRangeSelector/date-range-selector.ts
```diff
@@ -0,0 +1,168 @@
...
+        const currentDateTo = this.dateToPicker?.selectedDates;
+
+        return {
+            dateFrom: currentDateFrom ? currentDateFrom[0] : today,
+            dateTo: currentDateTo ? currentDateTo[0] : today,
+        };
+    }
+
+    private onSelectChange(event: CustomEvent) {
+        const value = event.detail.value as string;
```
fengelniederhammer, line 127:
```suggestion
    private onSelectChange(event: CustomEvent<{ value: string }>) {
        const value = event.detail.value;
```

fengelniederhammer (approved):
LGTM

PR #83, opened 2024-03-20
feat(components): add gs-date-range-changed event to date range selector

### LINKED ISSUE #80: Enable expect 'custom' for date range selector
In #9 we added a date range selector. One test types a specific date into the date from input element and tries to press 'enter'. This does work by hand but not programmatically. Therefor the assertion that the select changes to 'custom' can not be done in the test. 

Enable that assertion again, when hitting enter works. 

I could not find out what is causing this issue. I would either think of storybook or lit.

### LINKED ISSUE #81: Date range selector: Dispatch event

JonasKellerer:
Resolves #81, resolves #80

### Summary

This adds a event and a submit button to the date range selector. The event contains the dateFrom and dateTo of the selector.

### Screenshot

image

### PR Checklist

### DIFF
--- a/components/src/components/dateRangeSelector/date-range-selector.stories.ts
+++ b/components/src/components/dateRangeSelector/date-range-selector.stories.ts
@@ -1,12 +1,20 @@
 import { html } from 'lit';
 import { Meta, StoryObj } from '@storybook/web-components';
 import './date-range-selector';
-import { expect, userEvent, within } from '@storybook/test';
+import { expect, fn, userEvent, within } from '@storybook/test';
+import { withActions } from '@storybook/addon-actions/decorator';
+import { toYYYYMMDD } from './date-range-selector';
 
 const meta: Meta = {
     title: 'Input/Date Range Selector',
     component: 'gs-date-range-selector',
-    parameters: { fetchMock: {} },
+    parameters: {
+        fetchMock: {},
+        actions: {
+            handles: ['gs-date-range-changed'],
+        },
+    },
+    decorators: [withActions],
 };
 
 export default meta;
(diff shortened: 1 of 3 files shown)


components/src/components/dateRangeSelector/date-range-selector.ts
```diff
@@ -155,14 +147,45 @@ export class DateRangeSelector extends TailwindElement() {
...
     }
+
+    private submit() {
+        const dateFrom = this.dateFromPicker?.selectedDates[0];
+        const dateTo = this.dateToPicker?.selectedDates[0];
+
+        this.dispatchEvent(
+            new CustomEvent('gs-date-range-changed', {
+                detail: { dateFrom, dateTo },
+                bubbles: true,
```
fengelniederhammer, line 187: https://developer.mozilla.org/en-US/docs/Web/API/Event/composed

```suggestion
                bubbles: true,
                composed: true,
```

components/src/components/dateRangeSelector/date-range-selector.stories.ts
```diff
@@ -39,17 +46,29 @@ export const DateRangeSelectorWithSelectedRange: StoryObj = {
...
+            await expect(dateFrom()).toHaveValue(someDateInThePast);
+            await expect(canvas.getByRole('combobox')).toHaveValue('custom');
+        });
 
-            await expect(dateFrom).toHaveValue(someDateInThePast);
-            // TODO(#80): Reenable this line when hitting enter works on the input.
-            // await expect(canvas.getByRole('combobox')).toHaveValue('custom');
+        await step('Submit', async () => {
+            await userEvent.click(submitButton());
+            await expect(listenerMock).toHaveBeenCalled();
```
fengelniederhammer, line 71: I think it would also be good to check that the mock has been called with correct values.

components/src/components/dateRangeSelector/date-range-selector.ts
```diff
@@ -155,14 +147,45 @@ export class DateRangeSelector extends TailwindElement() {
...
                     class="input input-bordered rounded-none join-item"
                     type="text"
                     placeholder="Date to"
                     id="end-date-picker"
+                    @change=${(event: Event) => {
+                        event.preventDefault();
+                        this.dateFromPicker?.set('maxDate', this.dateToPicker?.selectedDates[0]);
+                        this.selectedValue = 'custom';
+                    }}
                 />
```
fengelniederhammer, line 166: Actually it would be nice if pressing enter in this field would also trigger the event.

components/src/components/dateRangeSelector/date-range-selector.ts
```diff
@@ -142,7 +134,7 @@ export class DateRangeSelector extends TailwindElement() {
 
     override render() {
         return html`
-            <div class="join">
+            <form class="join">
```
fengelniederhammer, line 137: As discussed: Either we use the submit of this form or we can revert it to a div.

components/src/components/dateRangeSelector/date-range-selector.ts
```diff
@@ -155,14 +147,45 @@ export class DateRangeSelector extends TailwindElement() {
                     type="text"
                     placeholder="Date from"
                     id="date-picker"
+                    @change=${(event: Event) => {
```
fengelniederhammer, line 150: As discussed: Selecting a new value here should also trigger the event.

fengelniederhammer (changes requested):

JonasKellerer:
As discussed we remove the submit button and submit, when the user
- hits enter
- selects one from the predetermined ranges
- leaves the component

JonasKellerer:
The component now looks like this without the submit button:
grafik

components/src/components/dateRangeSelector/date-range-selector.ts
```diff
@@ -155,14 +134,82 @@ export class DateRangeSelector extends TailwindElement() {
...
+        this.submit();
+    }
+
+    private submit() {
+        const dateFrom = this.dateFromPicker?.selectedDates[0];
+        const dateTo = this.dateToPicker?.selectedDates[0];
+
+        this.dispatchEvent(
+            new CustomEvent('gs-date-range-changed', {
+                detail: { dateFrom, dateTo },
```
fengelniederhammer, line 209: Ultimately this goes into LAPIS. So I think the event should contain dates that can be consumed by LAPIS, i.e. a string in the correct format (instead of a `Date` object). What do you think @chaoran-chen?

chaoran-chen: Yes, the output of the selectors should be a [`LapisFilter`](https://github.com/GenSpectrum/dashboards/blob/main/components/src/types.ts#L1).

fengelniederhammer (approved):
LGTM

PR #85, opened 2024-03-21
POC: Use Preact for render internal DOM

fengelniederhammer:
### Summary

This PR migrates "inner Lit components" to Preact compnents. Effectively we have two frameworks:
* Preact as a framework that renders graphs etc.
* Lit as a wrapper around the Preact islands that creates web components where necessary (i.e. only "top level components" will become web components)

### PR Checklist


JonasKellerer (approved):
LGTM

PR #89, opened 2024-03-26
feat(components): add mutation comparison with table

### LINKED ISSUE #77: Mutation comparison - Table: Add a table view that, for each mutation, shows the prevalence in the different variants

JonasKellerer:
resolves #77 

### Summary

This adds a new component, which shows a mutation comparison of variations.
Currently, there is only a table view.

### Screenshot

grafik

### PR Checklist


components/src/preact/mutationComparison/mutation-comparison.stories.tsx
```diff
@@ -0,0 +1,131 @@
...
+                lapisFilter: {
+                    country: 'Switzerland',
+                    pangoLineage: 'B.1.1.7',
+                    dateFrom: dateFromOtherVariant,
+                    dateTo: dateToOtherVariant,
+                },
+            },
+        ],
+        sequenceType: 'nucleotide',
+        views: ['table'],
```
fengelniederhammer, line 63: `meta` already has `views` set.

components/src/preact-components/display/mutation-comparison-component.stories.ts
```diff
@@ -0,0 +1,129 @@
...
+                lapisFilter: {
+                    country: 'Switzerland',
+                    pangoLineage: 'B.1.1.7',
+                    dateFrom,
+                    dateTo,
+                },
+            },
+        ],
+        sequenceType: 'nucleotide',
+        views: ['table'],
```
fengelniederhammer, line 66: Same here, I think we don't need it here, if `meta` already has it.

fengelniederhammer (approved):

chaoran-chen (commented):
I just looked at the screenshot and it's great!

One improvement for all tables: #91

components/src/preact/mutationComparison/mutation-comparison.stories.tsx
JonasKellerer: The views from meta is only to provide information, what can be selected in the controls. It does not set the value. One has to provide it as an arg.

Issue #88, opened 2024-03-26
PrevalenceOverTime - hide scaling selector on table


JonasKellerer:
Currently the scaling selector shows also on the table view. This should not be the case. For this you need to find a way to evaluate which is the active tab.

PR #96, opened 2024-04-02
58 allow users to set a min proportion

### LINKED ISSUE #58: Allow users to set a min proportion

JonasKellerer:
resolves #58

This is a draft and should show how the component could be used. The corresponding tests are still missing. 

### Summary

Adds a new component, which opens a dropdown. In this dropdown, users can select the min and max proportion.
Currently, this only applies to substitutions and deletions. Insertions stay untouched.

Currently, the tab switches back to the first tab when changing the proportion. This can only be fixed, once we can mark the current selected tab. This is similar to #88.

### Screenshot

grafik

### PR Checklist


components/src/preact/components/percent-intput.tsx
```diff
@@ -0,0 +1,46 @@
...
+    percentage: number;
+    setPercentage: (percentage: number) => void;
+};
+
+const percentageInRange = (percentage: number) => {
+    return percentage <= 100 && percentage >= 0;
+};
+
+export const PercentInput: FunctionComponent<PercentInputProps> = ({ percentage, setPercentage }) => {
+    const [error, setError] = useState(!percentageInRange(percentage));
```
fengelniederhammer, line 15:
```suggestion
    const [isError, setIsError] = useState(!percentageInRange(percentage));
```

components/src/preact/components/percent-intput.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const handleInputChange = (event: ChangeEvent<HTMLInputElement>) => {
+        const input = event.target as HTMLInputElement;
+        const value = Number(input.value);
+
+        const inRange = percentageInRange(value);
+
+        if (!inRange) {
+            setError(true);
+        } else {
+            setError(false);
```
fengelniederhammer, line 26: We also need to check this if `value` changes by changing the value via the slider:

* Input wrong value
* Input box gets read (as expected)
* change the value via the slider
* Input box stays red (should go back to "no error")

fengelniederhammer (commented):
Looks good in general :+1:

chaoran-chen:
Thanks, @JonasKellerer, it looks beautiful! Is it possible to enter small numbers? I.e., can we, for example, filter for mutations in the range of 0.01%-1%. That would be quite useful.

JonasKellerer:
Currently the precision is set to 3 digits, so 0.001% is possible. The up down arrows change the percentage by 0.1%. (I thought this is more often the use case). But I can also change this to 3 digits as well.

components/src/preact/components/proportion-selector.tsx
fengelniederhammer, line 1: image

By typing values into the inputs, I can generate invalid states. :)

components/src/preact/mutations/mutations.tsx
```diff
@@ -114,15 +118,26 @@ export const Mutations: FunctionComponent<MutationsProps> = ({ variant, sequence
...
+                return true;
+            }
+            return displayedSegments.some(
+                (displayedSegment) =>
+                    displayedSegment.segment === mutationEntry.mutation.segment && displayedSegment.checked,
+            );
+        });
+    };
+
+    const filterByProportion = (mutationEntry: MutationEntry[]) => {
```
fengelniederhammer, line 133:
```suggestion
    const filterByProportion = (mutationEntries: MutationEntry[]) => {
```

components/src/preact/mutations/mutations.tsx
fengelniederhammer, line 1: [video](https://github.com/GenSpectrum/dashboards/assets/92720311/8f8681e4-40e8-4d55-88bb-c7eb73435e6a)

When dragging one of those sliders, the table gets rerendered on every tick. Maybe we want to avoid that? But I also don't want to block on this.

fengelniederhammer (changes requested):

fengelniederhammer (approved):

components/src/preact/components/proportion-selector.tsx
JonasKellerer: As discussed we postpone this to later in this ticket: #115

components/src/preact/mutations/mutations.tsx
JonasKellerer: As discussed, this is intended behavior. Users should be able to see the changes immediately. (Until this takes up to much computing power.)

PR #98, opened 2024-04-02
feat(components): bring back the custom-elements-manifest analyzer

### LINKED ISSUE #34: Use custom-elements-manifest to document components
In #31 we removed the custom elements manifest because we did not use it and have no way to test it. Bring this back.

https://github.com/webcomponents/custom-elements-manifest

fengelniederhammer:
resolves #34

### Summary

I think IntelliJ has no proper support for the manifest, but the generated manifest looks ok (as far as I can judge). It isn't used anywhere, but it should probably be contained in the published package (once we publish).

### PR Checklist


components/package.json
```diff
@@ -11,6 +11,7 @@
         "lint": "npm run lint:lit-analyzer && npm run lint:eslint",
         "lint:eslint": "eslint 'src/**/*.{ts,tsx}'",
         "lint:lit-analyzer": "lit-analyzer",
+        "generate-manifest": "npx custom-elements-manifest analyze --litelement --globs src/**",
```
JonasKellerer, line 14: This seems to also provide definitions for all .ts files and also preact files. We only want the definitions for our lit components?

JonasKellerer (approved):
LGTM
(However, I dont see the code completion work yet with intelliJ)

PR #99, opened 2024-04-02
feat(components): use percent format for all table columns with prevalence

### LINKED ISSUE #91: Show proportions as percentages in tables
In all the tables, we should show proportions as `xx.xx%`. (In the downloaded CSV files, the values should **not** change.)

JonasKellerer:
resolves #91

### Summary

Use percent format for mutations, mutation-comparison and prevalecne-over-time.

### PR Checklist


components/src/preact/mutations/mutations-grid.tsx
```diff
@@ -159,6 +159,6 @@ export const MutationsGrid: FunctionComponent<MutationsGridProps> = ({ data, seq
     return <Table data={getTableData(data, sequenceType)} columns={getHeaders()} pagination={true} />;
 };
 
-const formatProportion = (proportion: number) => {
+export const formatProportion = (proportion: number) => {
```
fengelniederhammer, line 162: Let's move this function to a more central place? 

Maybe we can clean up a little. There are similar things in `components/charts/...`. We could move all that to some sort of "shared util" place (maybe with a better name)?

fengelniederhammer (approved):

PR #101, opened 2024-04-03
fix(components): move flaky date range test to playwright

JonasKellerer:
### Summary

The play function of storybook does not support pressing enter on the input form. This required to do for testing. Thus, we moved the test to playwright, where this is possible.

### PR Checklist


fengelniederhammer (approved):
LGTM

PR #104, opened 2024-04-04
check/test date granularity

### LINKED ISSUE #26: Check/Test date granularity
We have the component Temporal. This component is not much tested, but used very often. Add more tests. Especially, check why the dates with a daily granularity do not match the monthly granularity. (Both screenshots take the same input, but one of them goes from 2023-01 till 2024-01, the other only to 2024-11 and does not start from 2023-01)

grafik
grafik

fengelniederhammer:
resolves #26

### Summary

The actual issue described in the ticket, that the month range doesn't go far enough seems to be due to the smoothing window. I didn't observe any issues with that when reducing the smoothing window.

What I did though: 
* I added a couple of tests
* I fixed a bug in the sliding operator
* better naming for generic types.

### PR Checklist

### DIFF
--- a/components/src/operator/FillMissingOperator.ts
+++ b/components/src/operator/FillMissingOperator.ts
@@ -1,16 +1,16 @@
 import { Operator } from './Operator';
 import { Dataset } from './Dataset';
 
-export class FillMissingOperator<S, K extends keyof S> implements Operator<S> {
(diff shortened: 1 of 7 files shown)


JonasKellerer (approved):
Nice. Thank you.

PR #121, opened 2024-04-05
feat(components): mutation comparison: allow user to set min/max proportions

### LINKED ISSUE #108: Mutation Comparison: Allow users to set a min proportion
Similar to #58

fengelniederhammer:
resolves #108

### Summary

* In the Venn diagram, only consider a mutation of a variant, if its proportion is in the range.
* Only show the line of a mutation if the proportion of __any__ variant is in the range.

### Screenshot

grafik

### PR Checklist


components/src/preact/mutationComparison/mutation-comparison-venn.tsx
```diff
@@ -30,71 +31,93 @@ export const MutationComparisonVenn: FunctionComponent<MutationComparisonVennPro
...
+        if (divRef.current === null) {
+            return;
+        }
+        if (selectedDatasetIndex === null) {
+            divRef.current.innerText = noElementSelectedMessage;
+            return;
+        }
+
+        const values = sets.datasets[0].data[selectedDatasetIndex].values;
+        divRef.current!.innerText = `Mutations: ${values.join(', ')}` || '';
```
JonasKellerer, line 65: Maybe we can show the union on the label like: `Mutations [displayName1 ∪ displayName2]: A123T`. Then the user still knows what is displayed below, when the highlight changes?

components/src/operator/FetchMutationsOperator.ts
```diff
@@ -75,3 +75,7 @@ export type DeletionEntry = { type: 'deletion'; mutation: Deletion; count: numbe
 export type InsertionEntry = { type: 'insertion'; mutation: Insertion; count: number };
 
 export type MutationEntry = SubstitutionEntry | DeletionEntry | InsertionEntry;
+
+export function isSubstitutionOrDeletionEntry(entry: MutationEntry): entry is SubstitutionEntry | DeletionEntry {
```
JonasKellerer, line 79: This might become obsolete after #120

JonasKellerer (approved):
LGTM

components/src/preact/mutationComparison/mutation-comparison-venn.tsx
fengelniederhammer: I'll try it in a next PR :+1:

PR #132, opened 2024-04-10
feat(components): build and use components in a separate index.html

### LINKED ISSUE #35: Provide a demonstration page of our components
In #31 we used the `index.html` to show a demo version of the components. 

Put this demo to a more suitable place then in the root folder.

fengelniederhammer:
resolves #35

#### Uses vite to bundle the library
https://open-wc.org/guides/developing-components/publishing/ recommends that one should not bundle, but according to https://justinfagnani.com/2019/11/01/how-to-publish-web-components-to-npm/ the main point is that one should keep the dependencies external to the bundle. So this is what I ended up with:
* We need to preprocess the CSS (because of tailwind), so I figured it would be easiest to use vite.
  * I tried compiling with `tsc` and using `postcss` directly, but then Storybook didn't work anymore properly and I didn't want to go down the rabbit hole to fix that.
  * We already use vite to build for Storybook.
* Dependencies are kept external to the built library.
* We also provide `.d.ts` files and source maps, I didn't test whether they work though.

#### Sizes
```
dist/style.css                 94.15 kB │ gzip: 16.99 kB
dist/dashboard-components.js  213.34 kB │ gzip: 45.36 kB │ map: 291.88 kB
```

#### Other remarks

* Using `npm pack` and then install the packed tarball in an example seemed the closest-to-reality way of using our own library in an example to me.
* Currently, this only only creates ES modules. It's easy to extend this to create CJS modules, too: https://vitejs.dev/guide/build.html#library-mode. I'm not sure whether we need to/should support this. 
* This only builds a single entry point. I don't know whether we want to provide an entry point for every component.
  * The benefit would be that one could register components selectively.
  * It's also easy to change this later (and non-breaking), so I left it for the future.
* Consumers need to import the `style.css` in their HTML, because it sets CSS variables at `:root`. I think this needs to happen in the real document. When I tried it, it didn't work in the shadow DOM. Otherwise colors won't work, because themes (and thus colors) are set via those variables.
  * The CSS is contained twice (once in the code, once as `style.css`). This is room for improvement, because actually we only somehow need to set the color variables:
    * Either we somehow manage to strip the `style.css` to only contain the `:root` part,
    * or we manage to make it work via setting the colors on `:host`
      * I don't know whether that might work
      * The benefit would be that it doesn't need to be set globally.

### Screenshot

This is what the installed library contains in the example `node_modules`:
grafik

Somewhere I read that the published package should contain the code to make source maps work. I believe that also the `custom-elements.json` profits from that, since it references `src/...`.

### PR Checklist

  - `npm run build` in the CI
  - I'm unsure whether we somehow want to automatically test the example. For now this is a manual test. We could maybe try to build it without checking that it actually works.


components/src/utils/mutations.ts
```diff
@@ -59,14 +59,16 @@ export interface Mutation {
 }
 
 export class Substitution implements Mutation {
-    readonly code = `${this.segment ? `${this.segment}:` : ''}${this.valueAtReference}${this.position}${this.substitutionValue}`;
+    readonly code;
```
JonasKellerer, line 62: I like this change, but was there an issue with this in your PR?

examples/plainJavascript/index.html
```diff
@@ -0,0 +1,77 @@
...
+        />
+    </head>
+    <body>
+        <gs-app lapis="https://lapis.cov-spectrum.org/open/v1/sample">
+            <gs-location-filter
+                value="Europe / Switzerland"
+                fields='["region", "country", "division", "location"]'
+            ></gs-location-filter>
+            <gs-date-range-selector
+                selectedValue="last6Months"
```
JonasKellerer, line 23: I think this is not part of this PR, but when I change one date by hand I get the following error:
grafik

JonasKellerer (approved):
Nice. Now we can build on this.

components/src/utils/mutations.ts
fengelniederhammer: Yes, somehow the vite build complained about this. I don't know why exactly.

examples/plainJavascript/index.html
fengelniederhammer: Probably there is not data for the filter that you set (might also be that the code sets a wrong filter)

PR #135, opened 2024-04-10
feat(examples): add example React app to demonstrate how to use the components 

### LINKED ISSUE #133: Provide a simple React example to verify that the package can be used in JS code

fengelniederhammer:
resolves #133

### PR Checklist

### DIFF
--- a/examples/React/README.md
+++ b/examples/React/README.md
@@ -1,0 +1,16 @@
+# GenSpectrum components in React
+
+This demonstrates how to use GenSpectrum components in React.
+It was created using `npm create vite@latest React --template react-ts`.
+
+Install the packed version of the components:
+
+```bash
+npm run import-components
+```
+
+Start a test server to see the components in action:
+
+```bash
+npm run dev
+```
--- a/examples/React/src/App.tsx
+++ b/examples/React/src/App.tsx
@@ -1,0 +1,77 @@
+import {useEffect, useState} from 'react'
+import '@genspectrum/dashboard-components'
+import '@genspectrum/dashboard-components/style.css'
+
+function App() {
+    const [location, setLocation] = useState({
+        region: "Europe",
+        country: "Switzerland"
+    })
+    const [dateRange, setDateRange] = useState({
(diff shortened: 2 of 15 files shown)


examples/React/README.md
```diff
@@ -0,0 +1,17 @@
...
+Install the packed version of the components:
+
+```bash
+npm run import-components
+```
+
+Start a test server to see the components in action:
+
+```bash
+npm run start
```
JonasKellerer, line 14: This seems to be
```suggestion
npm run dev
```
?

JonasKellerer: How are the other commands used from the `package.json`?

examples/React/README.md
```diff
@@ -0,0 +1,17 @@
...
+```
+
+Then, open http://localhost:5173/ in your browser.
```
JonasKellerer, line 17: For me this opens in localhost:5174 ?

examples/React/src/App.tsx
```diff
@@ -0,0 +1,78 @@
...
+        if (locationFilter) {
+            locationFilter.addEventListener('gs-location-changed', handleLocationChange);
+        }
+
+        const dateRangeSelector = document.querySelector('gs-date-range-selector');
+        if (dateRangeSelector) {
+            dateRangeSelector.addEventListener('gs-date-range-changed', handleDateRangeChange);
+        }
+
+        // Cleanup function to remove event listeners
```
JonasKellerer, line 29: Do we need this comment?

JonasKellerer (approved):
LGTM

Only the README needs some care.

examples/React/README.md
fengelniederhammer: Maybe we should simply get rid of this line. vite tells you which url to open and it looks for a free port (if you already have another `vite dev` running that block port 5173)

examples/React/README.md
fengelniederhammer: They are not used. I deleted them.

Issue #129, opened 2024-04-09
Document components


fengelniederhammer:
Before publishing the npm package, think about how we can document the components:
* Storybook?
* README?
* Some separate md files?

fengelniederhammer:
We'll try to use Storybook for documentation (documenting components and general concepts, tutorials, etc.). Let's try it for one component and see whether it really fits.

fengelniederhammer:
It should be quite simple to write docs in `.mdx` files in Storybook: https://storybook.js.org/docs/writing-docs/mdx#setup-custom-documentation

PR #136, opened 2024-04-11
docs(components): annotate components that fire events

fengelniederhammer:
related to #129

### Summary

Provide JSDoc annotations and `HTMLElementEventMap` entries for the events that our input elements fire.
(and fix Storybook such that the "Actions" tab shows the events again)

### Screenshot

grafik

With the help of the `HTMLElementEventMap`, typescript can infer the type of the event in the React app:
grafik

### PR Checklist

### DIFF
--- a/examples/React/src/App.tsx
+++ b/examples/React/src/App.tsx
@@ -13,8 +13,11 @@ function App() {
     })
 
     useEffect(() => {
-        const handleLocationChange = (event: Event) => setLocation((event as CustomEvent).detail);
-        const handleDateRangeChange = (event: Event) => setDateRange((event as CustomEvent).detail);
+        const handleLocationChange = (event: CustomEvent) => setLocation(event.detail);
+        const handleDateRangeChange = (event: CustomEvent<{
+            dateFrom: string;
+            dateTo: string
+        }>) => setDateRange(event.detail);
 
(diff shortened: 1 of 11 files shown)


examples/React/src/App.tsx
```diff
@@ -13,8 +13,11 @@ function App() {
     })
 
     useEffect(() => {
-        const handleLocationChange = (event: Event) => setLocation((event as CustomEvent).detail);
-        const handleDateRangeChange = (event: Event) => setDateRange((event as CustomEvent).detail);
+        const handleLocationChange = (event: CustomEvent) => setLocation(event.detail);
+        const handleDateRangeChange = (event: CustomEvent<{
```
JonasKellerer, line 17: Why do we need the template annotation for the CustomEvent here, but not no the handleLocationChange?

JonasKellerer (approved):
LGTM

examples/React/src/App.tsx
fengelniederhammer: The location change event is a `CustomEvent<Record<string, string>>`, which isn't much better than `CustomEvent` which defaults to `CustomEvent<any>`. So TS doesn't complain here. We could however type it, since we know the fields that it might have...
For now this is only example code, so I didn't bother too much.

PR #137, opened 2024-04-11
feat(components): provide reference genome to all components

### LINKED ISSUE #92: Add reference genome to context
In many components we need the information if there are segments and which are those. To avoid to extract them every time, we decided to add the reference genome to the context in the GsApp. The reference genome can be queried from LAPIS directly.

When this is done, delete the special logic in `mutations.tsx` and `mutation-comparison.tsx`. Maybe there are more locations.

JonasKellerer:
resolves #92

### Summary

Provides the reference genome to all `gs-app` components through a context provider.
A general mock is introduced in `preview.ts`, since the reference genome is needed by all components which use `gs-app`. Specific values of the reference genome can be overwritten by targeted mocks in each story. 

### PR Checklist


components/src/preact/mutations/mutations.tsx
```diff
@@ -37,7 +39,7 @@ export interface MutationsProps {
 
 export const Mutations: FunctionComponent<MutationsProps> = ({ variant, sequenceType, views }) => {
     const lapis = useContext(LapisUrlContext);
-
+    const referenceGenome = useContext(ReferenceGenomeContext);
```
fengelniederhammer, line 42: How about moving this one layer down into `MutationsTabs`?
We could also take it a step further and wrap it in a `useSegments(sequenceType: SequenceType)`. 

Or we could also take it another step further (because that's how we currently exclusively use the segments):
```typescript
function useDisplayedSegments(sequenceType: SequenceType) {

    const displayedSegments = getSegmentNames(referenceGenome, sequenceType).map((segment) => ({
        segment,
        label: segment,
        checked: true,
    }));

    return useState<DisplayedSegment[]>(displayedSegments);
}
```

fengelniederhammer: (Same for mutation comparison)

fengelniederhammer (changes requested):
Looks good :+1: 

I think we should handle the case better when LAPIS fails (or is extremely slow) to deliver the reference genomes.

examples/React/src/App.tsx
fengelniederhammer, line 1: Looks like this has been formatted either with the wrong prettier settings or with IDE defaults. (Prettier is not installed in this example I think)

fengelniederhammer (commented):
The slot thingy seems to work. So let's try it like that.

examples/React/src/App.tsx
JonasKellerer: IntelliJ used its default settings. I found out, that it searches for the prettierrc in the current folder and all its parents. But you can not set it by hand...

components/src/web-components/app.ts
```diff
@@ -1,21 +1,42 @@
...
-        const children = [];
-        for (const childNode of this.childNodes) {
-            children.push(html`${childNode}`);
-        }
-        return html`${children}`;
+        return this.updateReferenceGenome.render({
+            complete: () => {
+                return html` <slot></slot>`;
+            },
+            error: () => html`<p>Error</p>`,
```
fengelniederhammer, line 35: As discussed, let's add a TODO

fengelniederhammer (approved):

components/src/web-components/app.ts
JonasKellerer: There is now this ticket https://github.com/GenSpectrum/dashboards/issues/143, and I added the TODO

PR #139, opened 2024-04-11
refactor(components): remove mutation cache

### LINKED ISSUE #123: Remove mutation cache
After #54.

Remove the mutation cache and use the new computer readable format of mutations and insertions directly.

JonasKellerer:
resolves #123

### Summary

Removes the mutation cache and uses the parsed mutations directly from LAPIS.

Also fixes wrongly assigned position to mock data.

### PR Checklist

### DIFF
--- a/components/src/operator/FetchInsertionsOperator.ts
+++ b/components/src/operator/FetchInsertionsOperator.ts
@@ -2,7 +2,7 @@ import { type Dataset } from './Dataset';
 import { type Operator } from './Operator';
 import { fetchInsertions } from '../lapisApi/lapisApi';
 import { type InsertionEntry, type LapisFilter, type SequenceType } from '../types';
-import { MutationCache } from '../utils/mutations';
+import { Insertion } from '../utils/mutations';
 
 export class FetchInsertionsOperator implements Operator<InsertionEntry> {
     constructor(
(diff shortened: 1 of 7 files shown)


components/src/operator/FetchInsertionsOperator.ts
```diff
@@ -2,28 +2,20 @@ import { type Dataset } from './Dataset';
...
-    private async fetchInsertions(
-        lapis: string,
-        signal?: AbortSignal,
-    ): Promise<{ insertion: string; count: number }[]> {
-        return (await fetchInsertions(lapis, { ...this.filter }, this.sequenceType, signal)).data;
-    }
-
     async evaluate(lapisUrl: string, signal?: AbortSignal): Promise<Dataset<InsertionEntry>> {
-        const insertions = await this.fetchInsertions(lapisUrl, signal);
+        const insertions = (await fetchInsertions(lapisUrl, { ...this.filter }, this.sequenceType, signal)).data;
```
fengelniederhammer, line 14: Why do we need the destructuring here?
```suggestion
        const insertions = (await fetchInsertions(lapisUrl, this.filter, this.sequenceType, signal)).data;
```

components/src/utils/mutations.spec.ts
```diff
@@ -1,22 +1,20 @@
-import { expect, describe, it } from 'vitest';
+import { describe, expect, it } from 'vitest';
 
-import { Deletion, Insertion, MutationCache, Substitution } from './mutations';
+import { Deletion, Insertion, Substitution } from './mutations';
 
 describe('MutationCache.getMutation', () => {
```
fengelniederhammer, line 5: Let's adapt the title ;)

components/src/utils/mutations.ts
```diff
@@ -125,9 +73,9 @@ export class Insertion implements Mutation {
     constructor(
         readonly segment: string | undefined,
         readonly position: number,
-        readonly value: string,
+        readonly insertedSymbols: string,
     ) {
-        this.code = `ins_${this.segment ? `${this.segment}:` : ''}${this.position}:${this.value}`;
+        this.code = `ins_${this.segment ? `${this.segment}:` : ''}${this.position}:${this.insertedSymbols}`;
```
fengelniederhammer, line 78: Should we also drag along the `insertion` field that LAPIS returns? This line could simply be replaced by using it. Same for the `mutation` field for mutations. That would save us this "assemble it back together" logic here.

fengelniederhammer (approved):

components/src/utils/mutations.spec.ts
fengelniederhammer: I'm also not sure whether we still need the `parse` methods that this tests - but I think we can delay removing it until we know what we need there, after doing the mutations filter field.

JonasKellerer: I would also wait on deleting it.

components/src/utils/mutations.ts
JonasKellerer: I would also wait on #8 here. When we use the parse function, we might have to create the string again on our own. I add this as a comment to #8

PR #148, opened 2024-04-17
build(root): lint commit messages

### LINKED ISSUE #147: lint commit messages
Enforce conventional commits since we rely on them for generating release changelogs.

fengelniederhammer:
resolves #147

### Summary

Using https://commitlint.js.org/.

I enforced that a scope is present in the message. If we notice that it feels too restrictive, we can remove that check later.

The last dependabot commit broke the `body-max-line-length` default of 100, so I set it to 200.

### PR Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "dashboards",
+  "private": true,
+  "version": "0.0.1",
+  "description": "A collection of GenSpectrum dashboards and components",
+  "type": "module",
+  "devDependencies": {
+    "@commitlint/cli": "^19.2.2",
+    "@commitlint/config-conventional": "^19.2.2",
+    "@commitlint/types": "^19.0.3"
+  },
+  "scripts": {
+    "commitlint:last-commit": "commitlint --from=HEAD~1 --verbose"
+  }
+}
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -1,0 +1,24 @@
+import {RuleConfigSeverity} from "@commitlint/types";
+
+/**
+ * @type {import('@commitlint/types').UserConfig}
+ */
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "dashboards",
+  "private": true,
+  "version": "0.0.1",
+  "description": "A collection of GenSpectrum dashboards",
```
JonasKellerer, line 5: Is this the right description? It seems to be the dashboards as well as the components.

package.json
```diff
@@ -0,0 +1,12 @@
...
+  "type": "module",
+  "devDependencies": {
+    "@commitlint/cli": "^19.2.2",
+    "@commitlint/config-conventional": "^19.2.2",
+    "@commitlint/types": "^19.0.3"
+  }
+}
```
JonasKellerer, line 12: Not this PR, but maybe we move our `prettierrc.json` also to top level. Then, intelliJ finds it in the example, as well as in the components folder, I think.

package.json
```diff
@@ -0,0 +1,12 @@
...
+  "name": "dashboards",
+  "private": true,
+  "version": "0.0.1",
+  "description": "A collection of GenSpectrum dashboards",
+  "type": "module",
+  "devDependencies": {
+    "@commitlint/cli": "^19.2.2",
+    "@commitlint/config-conventional": "^19.2.2",
+    "@commitlint/types": "^19.0.3"
+  }
```
JonasKellerer, line 11: Do we want to have a short action to run also locally? Or is this not possible?

package.json
fengelniederhammer: I also thought about that. Let's do it in a later PR :+1:

PR #152, opened 2024-04-17
fix(components): fix types of logit scale #144

### LINKED ISSUE #144: yAxisScaleType: complete types
When using `getYAxisScale`, we need to add a ts-expect-error. Add correct types for the function, so we don't need that any longer.

JonasKellerer:
resolves #144

### Summary

The logit scale is added to the ScaleTypeRegistry of chart.js. So the type is now correctly asserted.

### PR Checklist


fengelniederhammer (approved):
Nice!

PR #162, opened 2024-04-23
Start to document components

### LINKED ISSUE #129: Document components

fengelniederhammer:
resolves #129

### Summary

This is mostly meant to update the README as a preparation to publish a first version as an NPM package.

This also sketches how we can document components using Storybook. If you like this approach, we can create follow-up issues to document all existing components.

### Screenshot

This is the Storybook docs page for the location filter. Most parts of this are autogenerated from the `custom-elements.json`.
image

### PR Checklist

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -1,23 +1,108 @@
 # GenSpectrum components
 
-## Storybook
+This NPM package provides a collection of web components to build interactive dashboards that visualize
+data of a specific instance of [LAPIS](https://github.com/GenSpectrum/LAPIS).
+
+### Documentation
+
+Documentation of the components is available in the deployed [Storybook](https://genspectrum.github.io/dashboards/).
+
+We also provide examples of how to use the components in the [examples](../examples) folder.
+
+### npm
+
+To install the package, run:
+
+```bash
+npm i @genspectrum/dashboard-components
+```
+
+### Core Concepts
+
+Internally, the components use [Preact](https://preactjs.com/).
+We use [Lit](https://lit.dev/) to create web components.
+
+We primarily provide two kinds of components:
+
+-   **Visualization components** (charts, tables, etc.)
+    -   Those components fetch data from the LAPIS instance and visualize it.
+-   **Input components** that let you specify sequence filters for the LAPIS requests.
+    -   Input changes will fire events that can be listened to by the visualization components.
+        It is the responsibility of the dashbaord maintainer to listen to those events
+        and to wire the data correctly into the visualization components.
+
+## Local Development
+
+#### Installation
+
+```bash
+npm ci
+```
+
+### Custom Elements Manifest
+
+This package also ships a [Custom Elements Manifest](https://custom-elements-manifest.open-wc.org/),
+which is a formal specification of the web components.
+We use the `@custom-elements-manifest/analyzer` to autogenerate the `custom-elements.json` file:
+
+```bash
+npm run generate-manifest
+```
+
+or in watch mode:
+
+```bash
+npm run generate-manifest:watch
+```
+
+### Storybook
 
 We use Storybook to develop our components.
 
 To start Storybook, run:
 
 ```bash
+npm run storybook-preact
 npm run storybook
 ```
 
-Then, open http://localhost:6006/ in your browser.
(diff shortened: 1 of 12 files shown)


components/src/web-components/input/location-filter.mdx
fengelniederhammer, line 1: If we notice that there is a pattern across all components, we can also think about using a customized docs template: https://storybook.js.org/docs/writing-docs/doc-blocks#customizing-the-automatic-docs-page

components/README.md
```diff
@@ -1,23 +1,109 @@
...
+
+```bash
+npm ci
+```
+
+### Custom Elements Manifest
+
+This package also ships a [Custom Elements Manifest](https://custom-elements-manifest.open-wc.org/),
+which is a formal specification of the web components.
+We use the `@custom-elements-manifest/analyzer` to autogenerate the `custom-elements.json` file:
```
JonasKellerer, line 46: Maybe add, that one has to run the cem before looking at the docs in storybook. Otherwise they dont show.

components/src/web-components/input/location-filter-component.tsx
```diff
@@ -4,13 +4,42 @@ import { LocationFilter } from '../../preact/locationFilter/location-filter';
...
+ * ## Tag
+ *
+ * `gs-location-filter`
+ *
+ * ## Context
+ *
+ * This component provides an input field to specify filters for locations.
+ *
+ * It expects a list of fields that form a strict hierarchical order (such as continent, country, city).
+ * The component will fetch a list of all possible values from the Lapis instance.
```
JonasKellerer, line 16:
```suggestion
 * The component will retrieve a list of all possible values for these fields from the Lapis instance. This list is then utilized to display autocomplete suggestions and to validate the input.
 
Valid inputs must maintain a strict hierarchical order, beginning with the first field. Each element should be separated by a '/' sign. The input does not need to encompass the entire hierarchy and may terminate prematurely. In such cases, all missing elements are set to `undefined`.
``

File: components/src/web-components/input/location-filter-component.tsx [hunk 3]
```diff
@@ -4,13 +4,42 @@ import { LocationFilter } from '../../preact/locationFilter/location-filter';
...
+ *
+ * Given `fields` is `['field1', 'field2', ..., 'fieldN']`,
+ * then valid values for the location filter must be of the form `valueForField1 / valueForField2 / ... / valueForFieldK`,
+ * where `1 <= K <= N`.
+ * Values for the fields `i > K` are considered `undefined`.
+ *
+ * @fires {CustomEvent<Record<string, string>>} gs-location-changed
+ * Fired when the field was submitted with a valid location value.
+ * The `details` of this event contains an object with all `fields` as keys
+ * and the corresponding values as values, if they are not `undefined`.
```
JonasKellerer, line 26: Maybe add an example here like
```
{
continent: "Asia"
country: "China"
city: "Peking"
}
```

components/src/web-components/input/location-filter.mdx
```diff
@@ -0,0 +1,25 @@
+import { ArgTypes, Description, Meta, Story, Title, Source } from '@storybook/blocks';
+
+import * as LocationFilterStories from './location-filter-component.stories.ts';
```
JonasKellerer, line 3: Is the star notation here intentional?

JonasKellerer (approved):
Great start.
LGTM.

components/src/web-components/input/location-filter.mdx
fengelniederhammer: Yes. Storybook suggests to do it like that:
https://storybook.js.org/docs/writing-docs/mdx#basic-example

Probably those `of` props needs the whole module.

components/README.md
fengelniederhammer: Below, I put it like:
```bash
npm run generate-manifest
npm run storybook-preact
npm run storybook
```

So that should actually cover it. I added another sentence in normal language though.

But good point: We also need to do that before publishing the GitHub pages.

PR #163, opened 2024-04-24
chore(root,components): add AGPL 3 license

### LINKED ISSUE #161: Add license
This repository doesn't have a license yet. @chaoran-chen which one should we add?

fengelniederhammer:
resolves #161

### PR Checklist

### DIFF
--- a/components/package.json
+++ b/components/package.json
@@ -4,6 +4,7 @@
     "version": "0.0.1",
     "description": "GenSpectrum web components for building dashboards",
     "type": "module",
+    "license": "AGPL-3.0-only",
     "main": "dist/genspectrum-components.js",
     "module": "dist/genspectrum-components.js",
     "types": "dist/genspectrum-components.d.ts",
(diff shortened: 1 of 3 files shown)


components/package.json
```diff
@@ -4,6 +4,7 @@
     "version": "0.0.1",
     "description": "GenSpectrum web components for building dashboards",
     "type": "module",
+    "license": "AGPL-3.0-or-later",
```
fengelniederhammer, line 7: See https://docs.npmjs.com/cli/v10/configuring-npm/package-json#license

chaoran-chen: Where do you see the "or-later"? I can't find it on the page.

fengelniederhammer: https://spdx.org/licenses/ shows it. We could also use `-only`. I don't really know what the difference is.

chaoran-chen: Ah, interesting. Let's better fix it to `-only`.

fengelniederhammer: Done

PR #168, opened 2024-04-24
chore(components): make prettier ignore the CHANGELOG.md since it's autogenerated

fengelniederhammer:
resolves #

### PR Checklist

### DIFF
--- a/components/.prettierrc.js
+++ b/components/.prettierrc.js
@@ -1,0 +1,11 @@
+/** @type {import("prettier").Config} */
+const config = {
+    printWidth: 120,
+    tabWidth: 4,
+    trailingComma: 'all',
+    semi: true,
+    jsxSingleQuote: true,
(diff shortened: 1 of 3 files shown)


components/.prettierrc.js
fengelniederhammer, line 1: I like JS more than plain JSON files, thus this change.

components/.prettierrc.js
```diff
@@ -0,0 +1,11 @@
+/** @type {import("prettier").Config} */
```
JonasKellerer, line 1: This cant be a ts file?

JonasKellerer (approved):
LGTM

components/.prettierrc.js
fengelniederhammer: https://prettier.io/docs/en/configuration doesn't list TS, unfortunately.

PR #170, opened 2024-04-24
build(components): fix workflow steps after release created

fengelniederhammer:
### Summary

Apparently, `release_created` is only set when creating a release for the root of the repo. Instead, it seems that we can use `components--release_created`.
https://github.com/google-github-actions/release-please-action?tab=readme-ov-file#path-outputs

### PR Checklist


JonasKellerer (approved):
LGTM

PR #178, opened 2024-04-24
feat(components): new aggregate component #158

### LINKED ISSUE #158: Aggregated data table: Basic table
Given fields = ["country", "division", "host"], the table would like like

```
country | division | host | count | proportion
------------------------------------------------
Switzerland | Basel | Human | 12 | 0.4%
```

JonasKellerer:
resolves #158

### Summary

This PR introduces a new component, which shows the aggregated data from LAPIS as a table. The maintainer can select the fields to show and also add a filter. The data can be downloaded by users.

### Screenshot

Bildschirmfoto vom 2024-04-24 18-06-59

### PR Checklist


JonasKellerer:
I wasn't sure how to name the component, and also the heading. This can definitely be improved.

components/src/preact/aggregatedData/aggregate-table.tsx
```diff
@@ -0,0 +1,34 @@
...
+                name: 'count',
+                sort: true,
+            },
+            {
+                name: 'proportion',
+                sort: true,
+                formatter: (cell: number) => formatProportion(cell),
+            },
+        ];
+    };
```
fengelniederhammer, line 31: Does this actually need to be a function? Can't we just make it a normal variable? 

```suggestion
    const headers = [
            ...fields.map((field) => {
                return {
                    name: field,
                    sort: true,
                };
            }),
            {
                name: 'count',
                sort: true,
            },
            {
                name: 'proportion',
                sort: true,
                formatter: (cell: number) => formatProportion(cell),
            },
        ];
```

components/src/query/queryAggregateData.ts
fengelniederhammer, line 1: Let's add a little unit test for that? (Mocking LAPIS with msw)

components/src/web-components/display/aggregate-component.tsx
fengelniederhammer, line 1: This also needs to be exported in the `index.ts` (otherwise it won't be in the build output).

fengelniederhammer (approved):
LGTM

Are there any leftover tasks for which we need to create new issues for the parent epic? I can only think of the "Info button".

PR #180, opened 2024-04-25
refactor(components): use mutations directly in table instead of string representation 

### LINKED ISSUE #153: Refactor: use Mutations directly in tables
In tables, the mutation is stored as a string and then parsed. Use the mutation directly as data, and only print the human readable value.

JonasKellerer:
resolves #153

### Summary

Refactors the table for the mutations and mutation comparison, so that is uses the mutation type directly instead of its string representation.

### PR Checklist

### DIFF
--- a/components/src/preact/mutationComparison/getMutationComparisonTableData.ts
+++ b/components/src/preact/mutationComparison/getMutationComparisonTableData.ts
@@ -1,35 +1,53 @@
 import { type MutationData } from './queryMutationData';
 import { type Dataset } from '../../operator/Dataset';
+import { type Deletion, type Substitution } from '../../utils/mutations';
 import { type ProportionInterval } from '../components/proportion-selector';
 
 type Proportions = {
     [displayName: string]: number;
 };
 
+type MutationComparisonRow = {
+    mutation: Substitution | Deletion;
+    proportions: Proportions;
+};
+
 export function getMutationComparisonTableData(data: Dataset<MutationData>, proportionInterval: ProportionInterval) {
-    const mutationsToProportions = new Map<string, Proportions>();
+    const mutationsToProportions = new Map<string, MutationComparisonRow>();
 
     for (const mutationData of data.content) {
         for (const mutationEntry of mutationData.data) {
-            const mutation = mutationEntry.mutation.toString();
-            const proportions = mutationsToProportions.get(mutation) || {};
-            proportions[mutationData.displayName] = mutationEntry.proportion;
-            mutationsToProportions.set(mutation, proportions);
+            const mutationKey = mutationEntry.mutation.toString();
+            const existingRow = mutationsToProportions.get(mutationKey);
+
+            if (!existingRow) {
+                mutationsToProportions.set(
+                    mutationKey,
+                    initializeMutationRow(mutationEntry.mutation, mutationData.displayName, mutationEntry.proportion),
+                );
+            } else {
+                existingRow.proportions = updateProportions(
+                    existingRow.proportions,
+                    mutationData.displayName,
+                    mutationEntry.proportion,
+                );
+                mutationsToProportions.set(mutationKey, existingRow);
+            }
         }
     }
 
-    return [...mutationsToProportions.entries()]
-        .map(([mutation, proportions]) => {
+    return [...mutationsToProportions.values()]
+        .map((row) => {
             return {
-                mutation,
+                mutation: row.mutation,
                 ...data.content
                     .map((mutationData) => {
                         return {
-                            [`${mutationData.displayName} prevalence`]: proportions[mutationData.displayName] || 0,
+                            [`${mutationData.displayName} prevalence`]: row.proportions[mutationData.displayName] || 0,
                         };
                     })
                     .reduce((acc, val) => ({ ...acc, ...val }), {}),
-            } as { mutation: string } & Proportions;
+            } as { mutation: Substitution | Deletion } & Proportions;
         })
         .filter((row) =>
             Object.values(row).some(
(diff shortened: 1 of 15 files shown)


components/src/preact/mutationComparison/getMutationComparisonTableData.ts
```diff
@@ -1,35 +1,53 @@
...
+                mutation: row.mutation,
                 ...data.content
                     .map((mutationData) => {
                         return {
-                            [`${mutationData.displayName} prevalence`]: proportions[mutationData.displayName] || 0,
+                            [`${mutationData.displayName} prevalence`]: row.proportions[mutationData.displayName] || 0,
                         };
                     })
                     .reduce((acc, val) => ({ ...acc, ...val }), {}),
-            } as { mutation: string } & Proportions;
```
fengelniederhammer, line 32: I think we actually need some type annotation somewhere here. At least my IDE shows `{mutation: Substitution | Deletion}[]` as return type for `getMutationComparisonTableData`, which is not correct.

components/src/preact/mutationComparison/mutation-comparison-table.tsx
```diff
@@ -14,30 +15,29 @@ export interface MutationsTableProps {
...
-                    compare: (a: string, b: string) => {
-                        return sortSubstitutionsAndDeletions(a, b);
-                    },
+    const headers = [
+        {
+            name: 'Mutation',
+            sort: {
+                compare: (a: Substitution | Deletion, b: Substitution | Deletion) => {
+                    return sortSubstitutionsAndDeletions(a, b);
                 },
```
fengelniederhammer, line 24: A matter of taste, but this could be shortened:

```suggestion
                compare: sortSubstitutionsAndDeletions,
```

components/src/preact/mutationComparison/mutation-comparison.tsx
```diff
@@ -170,7 +170,16 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
...
                 className='mx-1 btn btn-xs'
-                getData={() => getMutationComparisonTableData({ content: filteredData }, proportionInterval)}
+                getData={() =>
+                    getMutationComparisonTableData({ content: filteredData }, proportionInterval).map((row) => {
+                        return {
+                            mutation: row.mutation.toString(),
+                            ...Object.entries(row)
+                                .filter(([key]) => key !== 'mutation')
+                                .reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {}),
+                        };
```
fengelniederhammer, line 180:
```suggestion
                        return {
                            ...row,
                            mutation: row.mutation.toString(),
                        };
```

I think this might actually do the same?

components/src/preact/mutations/getMutationsTableData.spec.ts
```diff
@@ -24,27 +24,17 @@ describe('getMutationsTableData', () => {
...
+        expect(result).toEqual(data);
     });
 
     test('should filter out data below/above proportionInterval', () => {
         const belowInterval = 0.03;
         const aboveInterval = 0.95;
         const inInterval = 0.5;
 
+        const substitutionInInterval = new Substitution('segment1', 'A', 'T', 123);
+        const deletionInInterval = new Deletion('segment2', 'C', 123);
```
fengelniederhammer, line 36: Those should probably be used somewhere in `data` below?

components/src/preact/mutations/getMutationsTableData.ts
```diff
@@ -8,7 +8,7 @@ export function getMutationsTableData(data: SubstitutionOrDeletionEntry[], propo
 
     return data.filter(byProportion).map((mutationEntry) => {
         return {
-            mutation: mutationEntry.mutation.toString(),
+            mutation: mutationEntry.mutation,
```
fengelniederhammer, line 11: Do we need this `map` step at all?

components/src/preact/mutations/mutations.tsx
```diff
@@ -174,15 +174,31 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
...
-                        getData={() => getMutationsTableData(filteredData.tableData, proportionInterval)}
+                        getData={() =>
+                            getMutationsTableData(filteredData.tableData, proportionInterval).map((row) => {
+                                return {
+                                    mutation: row.mutation.toString(),
+                                    type: row.type,
+                                    count: row.count,
+                                    proportion: row.proportion,
+                                };
+                            })
```
fengelniederhammer, line 185: We could also extend this so that `getData` accepts also record value that have a `toString` method. Then we wouldn't need this map step here.

components/src/preact/shared/sort/sortInsertions.ts
```diff
@@ -1,21 +1,14 @@
...
-        if (positionA !== positionB) {
-            return positionA - positionB;
-        }
-        return insertionA.insertedSymbols.localeCompare(insertionB.insertedSymbols);
+    if (a.position !== b.position) {
+        return comparePositions(a.position, b.position);
     }
-    throw new Error(`Invalid insertion: ${a} or ${b}`);
+
+    return a.insertedSymbols.localeCompare(b.insertedSymbols);
```
fengelniederhammer, line 13: Nice :)

components/src/preact/shared/sort/sortSubstitutionsAndDeletions.spec.ts
```diff
@@ -1,19 +1,28 @@
...
-        const a = 'A123A';
-        const b = 'A123B';
+        const a = new Substitution(undefined, 'A', 'A', 123);
+        const b = new Substitution(undefined, 'A', 'B', 123);
+
+        expect(sortSubstitutionsAndDeletions(a, b)).toBeLessThan(0);
+        expect(sortSubstitutionsAndDeletions(b, a)).toBeGreaterThan(0);
+    });
+
+    test('should sort for substituotionValue over deletion', () => {
```
fengelniederhammer, line 23:
```suggestion
    test('should sort for substitutionValue over deletion', () => {
```

fengelniederhammer (approved):
Looks good, just a couple of minor things :)

components/src/preact/shared/sort/sortInsertions.ts
JonasKellerer: :)

components/src/preact/mutations/getMutationsTableData.ts
JonasKellerer: I tried it out and the tests fail (because the table is not rendered. This is caused by gridjs. Gridjs uses the object and creates an array out of it. So the ordering of the elements is important; it should match the ordering of the headers)

PR #203, opened 2024-05-02
feat!(components): all input components should have a `initialValue` attribute

### LINKED ISSUE #190: Parameter to set the initial value
All input components should have a parameter that sets the initial value.

I think that the parameters should always start with `initial` to make clear that it will only affects the initial state.

fengelniederhammer:
resolves #190

### Summary

BREAKING CHANGE: rename `value` to `initialValue` in `gs-location-filter`

### PR Checklist

### DIFF
--- a/components/src/web-components/input/date-range-selector-component.stories.ts
+++ b/components/src/web-components/input/date-range-selector-component.stories.ts
@@ -4,13 +4,22 @@ import type { Meta, StoryObj } from '@storybook/web-components';
 import { html } from 'lit';
 
 import { LAPIS_URL } from '../../constants';
-import { type DateRangeSelectorProps } from '../../preact/dateRangeSelector/date-range-selector';
+import {
+    type DateRangeSelectorProps,
+    PRESET_VALUE_ALL_TIMES,
+    PRESET_VALUE_CUSTOM,
+    PRESET_VALUE_LAST_2_MONTHS,
+    PRESET_VALUE_LAST_2_WEEKS,
+    PRESET_VALUE_LAST_3_MONTHS,
+    PRESET_VALUE_LAST_6_MONTHS,
+    PRESET_VALUE_LAST_MONTH,
+} from '../../preact/dateRangeSelector/date-range-selector';
 import './date-range-selector-component';
 import '../app';
 import { toYYYYMMDD } from '../../preact/dateRangeSelector/dateConversion';
 import { withinShadowRoot } from '../withinShadowRoot.story';
 
-const meta: Meta<DateRangeSelectorProps> = {
+const meta: Meta<DateRangeSelectorProps<'CustomDateRange'>> = {
     title: 'Input/DateRangeSelector',
     component: 'gs-date-range-selector',
     parameters: {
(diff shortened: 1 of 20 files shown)


components/src/web-components/input/date-range-selector-component.tsx
```diff
@@ -1,27 +1,40 @@
...
 
 /**
  * @fires {CustomEvent<{ dateFrom: string; dateTo: string; }>} gs-date-range-changed - When the date range has changed
  */
 @customElement('gs-date-range-selector')
-export class DateRangeSelectorComponent extends PreactLitAdapter {
+export class DateRangeSelectorComponent<CustomLabel extends string> extends PreactLitAdapter {
     @property({ type: Array })
-    customSelectOptions: CustomSelectOption[] = [];
+    customSelectOptions: CustomSelectOption<CustomLabel>[] = [];
```
JonasKellerer, line 16: We might want to document somehow, that the template type of this component must match the provided labels of the "customSelectOptions".

components/src/web-components/input/date-range-selector-component.stories.ts
JonasKellerer, line 1: In storybook there is no handle to set the initial value.

components/src/web-components/input/location-filter-component.stories.ts
```diff
@@ -28,13 +28,16 @@ const Template: StoryObj<LocationFilterProps> = {
...
+                    .fields=${args.fields}
+                    initialValue=${ifDefined(args.initialValue)}
+                ></gs-location-filter>
             </div>
         </gs-app>`;
     },
     args: {
         fields: ['region', 'country', 'division', 'location'],
-        value: '',
+        initialValue: '',
```
JonasKellerer, line 40: When I set the initialValue in the story the component does not show the initial value.

components/src/web-components/input/text-input-component.tsx
JonasKellerer, line 1: In storybook there is no handle to set the initial value.

components/src/web-components/input/mutation-filter-component.stories.ts
```diff
@@ -22,18 +23,28 @@ const meta: Meta = {
...
     },
+    args: {
+        initialValue: [],
+    },
+};
+
+export const Default: StoryObj<MutationFilterProps> = {
+    ...Template,
+    args: {
+        initialValue: ['A123T'],
```
JonasKellerer, line 42: Same as for the location filter component. Changing this property does not change the showed component.

JonasKellerer (changes requested):
Nice changes. In the web components stories, some things are not working.

components/src/web-components/input/date-range-selector-component.tsx
fengelniederhammer: Certainly, but I would leave that to #199

components/src/web-components/input/location-filter-component.stories.ts
fengelniederhammer: The reason is the `useState` in the Preact component. It doesn't create a new component, but it leave the old one intact, which already has the state set. So the `initialValue` doesn't do anything if you change it on an already mounted component.

JonasKellerer: True. We might need to add this to the docu, so the user has to refresh the storybook to see the change. Otherwise one could think this is a bug.

fengelniederhammer: Is it enough to add this to the component docs?

JonasKellerer (approved):
LGTM

PR #207, opened 2024-05-07
feat!(components): set size of display components or resize to fit container #191

### LINKED ISSUE #191: Configurable size or responsive to parent
It should be possible to define the exact height and width of a component or make it responsive so that it fills the parent (both height- and width-wise).

JonasKellerer:
BREAKING_CHANGE: all display components have default width 100% and height 100vh

resolves #191

### Summary

Lets display components resize to the parent component and/or allow for more fine grained size setting.
Each component has now a size prop, which can set `width`, `height` and `minWidth` of the component. The sizes have to be provided in css format (eg. '100%', '300px', '30vh', ...). 
The tabs inside of one component have now the same size. If the content is larger than the provided space, a scroll bar is shown at the side.

### PR Checklist


fengelniederhammer:
Apparently our currentl commitlint config doesn't allow `feat!`.

examples/React/src/App.tsx
```diff
@@ -59,20 +59,33 @@ function App() {
...
-                views='["line", "table"]'
-            ></gs-prevalence-over-time>
-            <gs-prevalence-over-time
-                numerator={JSON.stringify(numerator)}
-                denominator={JSON.stringify(denominator)}
-                granularity='month'
-                smoothingWindow='0'
-                views='["bar", "table"]'
-            ></gs-prevalence-over-time>
+            <div style={{ display: 'flex', height: '300px', width: '300px' }}>
```
fengelniederhammer, line 62: The width of this div isn't respected by it's children, is it?

components/src/web-components/display/mutation-comparison-component.tsx
```diff
@@ -19,8 +20,18 @@ export class MutationComparisonComponent extends PreactLitAdapterWithGridJsStyle
     @property({ type: Array })
     views: View[] = ['table'];
 
+    @property({ type: Object })
+    size: Size | undefined = undefined;
```
fengelniederhammer, line 24: Please document this property.

(Also for all other components)

components/src/web-components/display/aggregate-component.stories.ts
```diff
@@ -49,6 +49,7 @@ export const Table: StoryObj<AggregateProps> = {
                     .fields=${args.fields}
                     .filter=${args.filter}
                     .views=${args.views}
+                    .size=${{ width: '100%', height: '70vh' }}
```
fengelniederhammer, line 52: Let's maybe also have this in the Storybook `args`?

components/src/preact/aggregatedData/aggregate.stories.tsx
```diff
@@ -37,11 +37,9 @@ export default meta;
 
 export const Default: StoryObj<AggregateProps> = {
     render: (args) => (
-        <div class='max-w-screen-lg'>
-            <LapisUrlContext.Provider value={LAPIS_URL}>
-                <Aggregate {...args} />
-            </LapisUrlContext.Provider>
-        </div>
+        <LapisUrlContext.Provider value={LAPIS_URL}>
+            <Aggregate {...args} size={{ width: '100%', height: '70vh' }} />
```
fengelniederhammer, line 41: Let's maybe also have `size` in the Storybook `args`?

components/src/web-components/display/mutation-comparison-component.stories.ts
```diff
@@ -30,15 +30,14 @@ export default meta;
...
-                    .views=${args.views}
-                ></gs-mutation-comparison-component>
-            </gs-app>
-        </div>
+        <gs-app lapis="${LAPIS_URL}">
+            <gs-mutation-comparison-component
+                .variants=${args.variants}
+                .sequenceType=${args.sequenceType}
+                .views=${args.views}
+                .size=${{ width: '100%', height: '50vh' }}
```
fengelniederhammer, line 38: Can we use a larger height here? That would make the Venn diagram larger, so that it takes a bigger portion of the total pixels of the screenshot.

components/src/preact/components/tabs.tsx
```diff
@@ -14,34 +14,58 @@ interface ComponentTabsProps {
...
-                </div>
-            </Fragment>
-        );
-    });
+    useEffect(() => {
+        if (tabRef.current) {
+            const heightOfTabs = tabRef.current.getBoundingClientRect().height;
+            setHeightOfTabs(`${heightOfTabs}px`);
+        }
+    }, []);
```
fengelniederhammer, line 25: Is it enough that this only runs once? Or would this need to rerun on some sort of window resize?

components/src/preact/components/tabs.tsx
```diff
@@ -14,34 +14,58 @@ interface ComponentTabsProps {
...
+                className={`p-2 border-2 border-gray-100 rounded-b-md rounded-tr-md ${activeTab === tabs[0].title ? '' : 'rounded-tl-md'}`}
+                style={{ height: `calc(100% - ${heightOfTabs})` }}
+            >
+                {tabs.map((tab) => (
+                    <div className='h-full overflow-auto' key={tab.title} hidden={activeTab !== tab.title}>
+                        {tab.content}
+                    </div>
+                ))}
+            </div>
+            <div className='tabs tabs-lifted' />
```
fengelniederhammer, line 68: Is this line necessary?

fengelniederhammer (commented):
Looks good :+1:

fengelniederhammer:
Probably at some point a tutorial in the docs "How to style our components" would be nice.

examples/React/src/App.tsx
JonasKellerer: Hm... It should've. The `width: '50%'` should be relative. I have a look into it.

components/src/preact/components/resize-container.tsx
```diff
@@ -0,0 +1,38 @@
...
+    defaultSize?: Size;
+}
+
+export const ResizeContainer: FunctionComponent<ResizeContainerProps> = ({ children, size, defaultSize }) => {
+    return <div style={extendByDefault(size, defaultSize)}>{children}</div>;
+};
+
+export const basicDefaultSize = {
+    width: '100%',
+    height: '100vh',
```
JonasKellerer, line 20: As discussed: we dont set the default height here, since this should be per component.

examples/React/src/App.tsx
JonasKellerer: I found it. Apparently, flex overwrites the width of a div. This way the child component no longer can reference the width of that component. The solution is to either put the component into its own div with a set size, or set the size of the component directly through the size prop. I added a corresponding example.

components/src/preact/components/tabs.tsx
JonasKellerer: I think it is enough to run this once. This provides the size in pixel. The height of the tabs should only change, when the size of the text changes, or if elements on the toolbar change their size. However, we dont want the latter.

fengelniederhammer (approved):
LGTM, let's try it.

Issue #210, opened 2024-05-08
Unify component tags


fengelniederhammer:
Currently we have components with custom tags `gs-...-component`, such as `gs-mutation-comparison-component`, and components with custom tags `gs-...`, such as `gs-prevalence-over-time`.

Unify this. Either all components should end with `-component`, or they should not have the `-component` part in their tag. I prefer the version **without** `-component`, because IMO the suffix provides to additional value to the tag name.

PR #212, opened 2024-05-08
docs(components): document `gs-relative-growth-advantage`

### LINKED ISSUE #198: Document `gs-relative-growth-advantage`

fengelniederhammer:
resolves #198

### PR Checklist

### DIFF
--- a/components/src/web-components/display/relative-growth-advantage-component.stories.ts
+++ b/components/src/web-components/display/relative-growth-advantage-component.stories.ts
@@ -3,10 +3,19 @@ import { html } from 'lit';
 
 import './relative-growth-advantage-component';
 import '../app';
+import { withComponentDocs } from '../../../.storybook/ComponentDocsBlock';
 import { AGGREGATED_ENDPOINT, LAPIS_URL } from '../../constants';
 import denominator from '../../preact/relativeGrowthAdvantage/__mockData__/denominator.json';
 import numerator from '../../preact/relativeGrowthAdvantage/__mockData__/numerator.json';
 import { type RelativeGrowthAdvantageProps } from '../../preact/relativeGrowthAdvantage/relative-growth-advantage';
+
+const codeExample = String.raw`
+<gs-relative-growth-advantage
+    numerator='{ "country": "Switzerland", "pangoLineage": "B.1.1.7", "dateFrom": "2020-12-01" }'
+    denominator='{ "country": "Switzerland", "dateFrom": "2020-12-01" }'
+    generationTime="7"
+    views='["line"]'
+></gs-relative-growth-advantage>`;
 
 const meta: Meta<RelativeGrowthAdvantageProps> = {
     title: 'Visualization/Relative growth advantage',
@@ -21,6 +30,15 @@ const meta: Meta<RelativeGrowthAdvantageProps> = {
         },
         size: [{ control: 'object' }],
     },
+    parameters: withComponentDocs({
+        componentDocs: {
+            tag: 'gs-relative-growth-advantage',
+            opensShadowDom: true,
+            expectsChildren: false,
+            codeExample,
+        },
+    }),
+    tags: ['autodocs'],
 };
 
 export default meta;
--- a/components/src/web-components/display/relative-growth-advantage-component.tsx
+++ b/components/src/web-components/display/relative-growth-advantage-component.tsx
@@ -2,21 +2,59 @@ import { customElement, property } from 'lit/decorators.js';
 
 import { RelativeGrowthAdvantage, type View } from '../../preact/relativeGrowthAdvantage/relative-growth-advantage';
 import type { LapisFilter } from '../../types';
+import { type Equals, type Expect } from '../../utils/typeAssertions';
 import { PreactLitAdapter } from '../PreactLitAdapter';
 
+/**
+ * We assume a discrete time model, where new infections happen exactly every `generationTime` days.
+ * This is what we call a "generation".
+ *
+ * This component estimates the relative growth advantage of a variant by performing a logistic regression.
+ * Based on the inferred logistic growth rate, we derive the relative growth advantage (per generation).
+ *
+ * For details on the scientific method, see:
+ * Chen, Chaoran, et al. "Quantification of the spread of SARS-CoV-2 variant B.1.1.7 in Switzerland." Epidemics (2021);
+ * doi: [10.1016/j.epidem.2021.100480](https://doi.org/10.1016/j.epidem.2021.100480)
+ *
+ * This component fetches aggregated data from LAPIS.
+ * Then the data is sent to `https://cov-spectrum.org/api/v2/computed/model/chen2021Fitness`
+ * which performs the logistic regression and calculates the relative growth advantage.
+ *
+ * ## Views
+ *
+ * ### Line View
+ *
+ * The line view shows the relative growth advantage over time in a line chart.
+ * The dots in the plot show the proportions of the focal variant (`numerator`) to the `denominator` variant
+ * for every day as observed in the data.
+ * The line shows a logistic curve fitted to the data points, including a 95% confidence interval.
+ */
 @customElement('gs-relative-growth-advantage')
 export class RelativeGrowthAdvantageComponent extends PreactLitAdapter {
+    /**
+     * The LAPIS filter for the focal variant.
+     */
     @property({ type: Object })
-    numerator: LapisFilter = {};
+    numerator: Record<string, string | number | null | boolean> = {};
 
+    /**
+     * The LAPIS filter for the variant that the focal variant (`numerator`) should be compared to.
+     */
     @property({ type: Object })
-    denominator: LapisFilter = {};
+    denominator: Record<string, string | number | null | boolean> = {};
 
+    /**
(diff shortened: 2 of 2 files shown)


components/src/web-components/display/relative-growth-advantage-component.tsx
```diff
@@ -2,21 +2,46 @@ import { customElement, property } from 'lit/decorators.js';
...
+     */
     @property({ type: Object })
-    denominator: LapisFilter = {};
+    denominator: Record<string, string | number | null | boolean> = {};
 
+    /**
+     * TODO: What does this do?
+     */
     @property({ type: Number })
     generationTime: number = 7;
```
fengelniederhammer, line 38: @chaoran-chen What does the `generationTime` do?

components/src/web-components/display/relative-growth-advantage-component.tsx
```diff
@@ -2,21 +2,46 @@ import { customElement, property } from 'lit/decorators.js';
 
 import { RelativeGrowthAdvantage, type View } from '../../preact/relativeGrowthAdvantage/relative-growth-advantage';
 import type { LapisFilter } from '../../types';
+import { type Equals, type Expect } from '../../utils/typeAssertions';
 import { PreactLitAdapter } from '../PreactLitAdapter';
 
+/**
+ * TODO: What do we see here?
```
fengelniederhammer, line 9: @chaoran-chen What does the relative growth advantage actually show? How is it different from the prevalence over time?

components/src/web-components/display/relative-growth-advantage-component.tsx
```diff
@@ -2,21 +2,46 @@ import { customElement, property } from 'lit/decorators.js';
...
+ *
+ * This component fetches aggregated data from LAPIS.
+ * Then the data is sent to `https://cov-spectrum.org/api/v2/computed/model/chen2021Fitness` to compute TODO: WHAT?
```
fengelniederhammer, line 12: @chaoran-chen What does this endpoint actually do? I think it's at least worth mentioning that there is another service involved.

components/src/web-components/display/relative-growth-advantage-component.tsx
chaoran-chen: This generation time represents the number of days over which we measure the variant's relative growth advantage. For example, if we set a generation time of 7 days, then we estimate the growth advantage per week.

components/src/web-components/display/relative-growth-advantage-component.tsx
chaoran-chen: The component estimates the relative growth advantage of a variant by performing a logistic regression. The dots in the plot show the proportions of the focal variant for every day as observed in the data. The line shows a logistic curve fitted to the data points. Based on the inferred logistic growth rate, we derive the relative growth advantage per generation time.

components/src/web-components/display/relative-growth-advantage-component.tsx
chaoran-chen: The endpoint performs a logistic regression on the provided data points and calculates the relative growth advantage (per generation).

fengelniederhammer: Should we create an issue somewhere for this as a reminder? Or do we just keep it in mind?

chaoran-chen: I think that we can just keep it in mind :)

components/src/web-components/display/relative-growth-advantage-component.tsx
fengelniederhammer, line 1: Reminder: Squash before merge

components/src/web-components/display/relative-growth-advantage-component.tsx
```diff
@@ -2,21 +2,52 @@ import { customElement, property } from 'lit/decorators.js';
...
+ * which performs the logistic regression and calculates the relative growth advantage.
+ *
+ * ## Views
+ *
+ * ### Line View
+ *
+ * The line view shows the relative growth advantage over time in a line chart.
+ * The dots in the plot show the proportions of the focal variant (`numerator`) to the `denominator` variant
+ * for every day as observed in the data.
+ * The line shows a logistic curve fitted to the data points, including a confidence interval.
```
fengelniederhammer, line 23: Do we need to add more details on the confidence interval here (like the method how it was computed)?

components/src/web-components/display/relative-growth-advantage-component.tsx
```diff
@@ -2,21 +2,52 @@ import { customElement, property } from 'lit/decorators.js';
 
 import { RelativeGrowthAdvantage, type View } from '../../preact/relativeGrowthAdvantage/relative-growth-advantage';
 import type { LapisFilter } from '../../types';
+import { type Equals, type Expect } from '../../utils/typeAssertions';
 import { PreactLitAdapter } from '../PreactLitAdapter';
 
+/**
+ * This component estimates the relative growth advantage of a variant by performing a logistic regression.
+ * Based on the inferred logistic growth rate, we derive the relative growth advantage (per generation).
```
fengelniederhammer, line 10: Do we need a better explanation what "per generation" means here?

components/src/web-components/display/relative-growth-advantage-component.tsx
chaoran-chen: Maybe write "95% confidence interval", that should be enough for now/this place I think. Details can be found in the publication

> Chen, Chaoran, et al. "Quantification of the spread of SARS-CoV-2 variant B.1.1.7 in Switzerland." Epidemics (2021); doi: [10.1016/j.epidem.2021.100480](https://doi.org/10.1016/j.epidem.2021.100480)

We can add a reference in the docs.

components/src/web-components/display/relative-growth-advantage-component.tsx
chaoran-chen: Hm.. Do you have a suggestion?

components/src/web-components/display/relative-growth-advantage-component.stories.ts
```diff
@@ -20,6 +29,15 @@ const meta: Meta<RelativeGrowthAdvantageProps> = {
             control: { type: 'check' },
         },
     },
+    parameters: withComponentDocs({
+        componentDocs: {
+            tag: 'gs-relative-growth-advantage',
```
JonasKellerer, line 34:
```suggestion
            tag: 'gs-relative-growth-advantage-component',
```

components/src/web-components/display/relative-growth-advantage-component.tsx
```diff
@@ -2,21 +2,52 @@ import { customElement, property } from 'lit/decorators.js';
...
+    /**
+     * The LAPIS filter for the variant that the focal variant (`numerator`) should be compared to.
+     */
     @property({ type: Object })
-    denominator: LapisFilter = {};
+    denominator: Record<string, string | number | null | boolean> = {};
 
+    /**
+     * The generation time represents the number of days over which the variant's relative growth advantage is measured.
+     * For example, if we set a generation time of 7 days, then we estimate the growth advantage per week.
```
JonasKellerer, line 41: Only as a question: is this the same as the smoothing window?

JonasKellerer (approved):
LGTM

components/src/web-components/display/relative-growth-advantage-component.tsx
fengelniederhammer: > Only as a question: is this the same as the smoothing window?
@chaoran-chen, is it?

components/src/web-components/display/relative-growth-advantage-component.stories.ts
fengelniederhammer: Actually, this doesn't have `-component` in its name. See #210

components/src/web-components/display/relative-growth-advantage-component.tsx
fengelniederhammer: Not really. What is a generation in this case? Is there a generation for every day (smoothed over the `generationTime`?

components/src/web-components/display/relative-growth-advantage-component.tsx
chaoran-chen: No, it's not smoothing. The original idea and why it is called generation time $g$ is that if we assume that we have new infections exactly every $g$ days (in case of a discrete time model which we assume here) and that there are no differences between variants. Then, we are calculating how much faster a variant grows compared to another variant in each generation.

However, the infection times is not very clear and can change for different variants. Thus, we had the issue that different people, when they calculated the growth advantage, used different generation times and the numbers were not comparable. Therefore, we now set the generation time usually to 7 days which gives us how much more a variant can grow within a week.

Mathematically, we calculate the logistic growth rate $a$ and then derive the fitness/growth advantage through $f = e^{ag} - 1$.

components/src/web-components/display/relative-growth-advantage-component.tsx
chaoran-chen: (See [here](https://github.com/GenSpectrum/dashboards/pull/212#discussion_r1599511641))

PR #225, opened 2024-05-15
feat(components): add error boundary to components

### LINKED ISSUE #143: Components: Develop error handling
Develop a method to provide the user with error information. Most of the time error occur when LAPIS is not reachable. This can already happen in the `gs-app` component fetching the reference genome.

More tbd

JonasKellerer:
resolves #143

### Summary

Adds a error boundary to each component. This will catch every error thrown in its children. If the error is a `UserFacingError` then the content of the error message will also be shown. Otherwise it will be omitted. The error should take up as much space as the original component. 

This also introduces a loading skeleton for those components, which have this state.

When the reference genome can not be fetched, an error banner will be shown.

### Screenshot

grafik

Error banner when error while fetching reference genome.
grafik

### PR Checklist

### DIFF
--- a/components/src/preact/components/error-boundary.stories.tsx
+++ b/components/src/preact/components/error-boundary.stories.tsx
@@ -1,0 +1,62 @@
+import { type Meta, type StoryObj } from '@storybook/preact';
+import { expect, waitFor, within } from '@storybook/test';
+
+import { ErrorBoundary } from './error-boundary';
+
+const meta: Meta = {
+    title: 'Component/Error boundary',
+    component: ErrorBoundary,
+    parameters: { fetchMock: {} },
+    argTypes: {
+        size: { control: 'object' },
+        defaultSize: { control: 'object' },
+        headline: { control: 'text' },
+    },
+};
+
+export default meta;
+
+export const ErrorBoundaryWithoutErrorStory: StoryObj = {
+    render: (args) => (
+        <ErrorBoundary size={args.size} defaultSize={args.defaultSize} headline={args.headline}>
+            <div>Some content</div>
+        </ErrorBoundary>
+    ),
+    args: {
+        size: { height: '600px', width: '100%' },
+        defaultSize: { height: '600px', width: '100%' },
+        headline: 'Some headline',
+    },
+
+    play: async ({ canvasElement }) => {
+        const canvas = within(canvasElement);
+        const content = canvas.getByText('Some content', { exact: false });
+        await waitFor(() => expect(content).toBeInTheDocument());
+        await waitFor(() => expect(canvas.queryByText('Some headline')).not.toBeInTheDocument());
+    },
+};
+
+export const ErrorBoundaryWithErrorStory: StoryObj = {
+    render: (args) => (
+        <ErrorBoundary size={args.size} defaultSize={args.defaultSize} headline={args.headline}>
+            <ContentThatThrowsError />
+        </ErrorBoundary>
+    ),
+    args: {
+        size: { height: '600px', width: '100%' },
+        defaultSize: { height: '600px', width: '100%' },
+        headline: 'Some headline',
+    },
+
+    play: async ({ canvasElement }) => {
+        const canvas = within(canvasElement);
+        const content = canvas.queryByText('Some content.', { exact: false });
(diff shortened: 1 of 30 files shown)


components/src/preact/components/error-display.tsx
```diff
@@ -1,5 +1,20 @@
...
+
 export const ErrorDisplay: FunctionComponent<{ error: Error }> = ({ error }) => {
-    return <div>Error: {error.message}</div>;
+    const errorTypesToShow = ['UserFacingError'];
+
+    return (
+        <div className='h-full w-full rounded-md border-2 border-gray-100 p-2 flex items-center justify-center flex-col'>
+            <div className='text-red-700 font-bold'>Error</div>
+            <div>Oops! Something went wrong.</div>
+            {errorTypesToShow.includes(error.name) && <div className='text-sm text-gray-600'>{error.message}</div>}
```
fengelniederhammer, line 17:
```suggestion
            {error instanceof UserFacingError && <div className='text-sm text-gray-600'>{error.message}</div>}
```
Why don't we simply do this?

components/src/preact/components/error-boundary.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+        await waitFor(() => expect(canvas.getByText('Some headline')).toBeInTheDocument());
+        await waitFor(() => expect(canvas.getByText('Error')).toBeInTheDocument());
+    },
+};
+
+const ContentThatThrowsError = () => {
+    throw new Error('Some error');
+
+    // eslint-disable-next-line no-unreachable
+    return <div>Some content</div>;
```
fengelniederhammer, line 64:
```suggestion
```

components/src/preact/mutations/mutations.tsx
```diff
@@ -30,57 +31,58 @@ import { useQuery } from '../useQuery';
...
     variant: LapisFilter;
     sequenceType: SequenceType;
     views: View[];
     size?: Size;
     headline?: string;
 }
 
+export interface MutationsProps extends MutationsInnerProps {
+    size?: Size;
+    headline?: string;
```
fengelniederhammer, line 44: Aren't they already in the `MutationsInnerProps`?

components/src/web-components/app.ts
```diff
@@ -55,7 +55,10 @@ export class App extends LitElement {
     override render() {
         return this.updateReferenceGenome.render({
             complete: () => html` <slot></slot>`,
-            error: () => html`<p>Error</p>`, // TODO(#143): Add more advanced error handling
+            error: () =>
+                html` <div class="m-2 w-full alert alert-error">
+                    Error: Can not fetch reference genome. Is LAPIS available?
```
fengelniederhammer, line 60:
```suggestion
                    Error: Cannot fetch reference genome. Is LAPIS available?
```

components/src/web-components/app.ts
```diff
@@ -55,7 +55,10 @@ export class App extends LitElement {
...
+                </div>`,
             pending: () => html` <p>Loading reference genomes...</p> `,
```
fengelniederhammer, line 62: I wonder whether we should show anything int the pending case... This looks a bit weird.

components/src/preact/components/error-display.tsx
JonasKellerer: True, but I wanted this to be extensible, so we may define many different UserFacingError types in the future.

fengelniederhammer: I don't know whether that is the right way. If you want to extend it, you can still create a new child class of `UserFacingError`.

components/src/preact/mutations/mutations.tsx
JonasKellerer: True, they are leftover. I deleted them.

components/src/web-components/app.ts
JonasKellerer: Good question. I am not entirely sure if we even want to display something. This will shift the layout for the user, every time the reference genome is loaded. What do you think?

JonasKellerer:
As discussed we log the error message for non user facing errors in the console.

PR #242, opened 2024-05-16
feat(components): move width and height outside of size property

### LINKED ISSUE #237: Component sizing: `width` and `height` atrributes
Instead of passing the size as an object to the `size` attribute, give the components two attributes `width` and `height`.

JonasKellerer:
BREAKING CHANGE: remove size property from components and use width and height 

resolves #237

### PR Checklist

### DIFF
--- a/components/src/web-components/input/gs-mutation-filter.tsx
+++ b/components/src/web-components/input/gs-mutation-filter.tsx
@@ -75,29 +75,34 @@ export class MutationFilterComponent extends PreactLitAdapter {
     @property()
     initialValue:
         {
-              nucleotideMutations: string[];
-              aminoAcidMutations: string[];
-              nucleotideInsertions: string[];
-              aminoAcidInsertions: string[];
-          }
+            nucleotideMutations: string[];
+            aminoAcidMutations: string[];
+            nucleotideInsertions: string[];
+            aminoAcidInsertions: string[];
+        }
         | string[]
         | undefined = undefined;
 
     /**
-     * The size of the component.
+     * The width of the component.
      *
-     * If not set, the component will take the full width of its container with height 700px.
-     *
-     * The width and height should be a string with a unit in css style, e.g. '100%', '500px' or '50vh'.
-     * If the unit is %, the size will be relative to the container of the component.
+     * Visit https://genspectrum.github.io/dashboards/?path=/docs/components-size-of-components--docs for more information.
      */
-    @property({ type: Object })
-    size: { width?: string; height?: string } | undefined = undefined;
+    @property({ type: String })
+    width: string = '100%';
+
+    /**
+     * The height of the component.
+     *
+     * Visit https://genspectrum.github.io/dashboards/?path=/docs/components-size-of-components--docs for more information.
+     */
+    @property({ type: String })
+    height: string = '6.5rem';
 
     override render() {
         return (
             <ReferenceGenomesAwaiter>
-                <MutationFilter initialValue={this.initialValue} size={this.size} />
+                <MutationFilter initialValue={this.initialValue} width={this.width} height={this.height} />
             </ReferenceGenomesAwaiter>
         );
     }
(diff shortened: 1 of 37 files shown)


components/src/web-components/input/gs-mutation-filter.tsx
```diff
@@ -68,29 +68,40 @@ export class MutationFilterComponent extends PreactLitAdapter {
...
-    size: { width?: string; height?: string } | undefined = undefined;
+    width: string | undefined = undefined;
+
+    /**
+     * The height of the component.
+     *
+     * If not set, the component will use the default height of 700px.
+     *
+     * The height should be a string with a unit in css style, e.g. '100%', '500px' or '50vw'.
+     * If the unit is %, the size will be relative to the container of the component.
```
fengelniederhammer, line 96: Maybe we can simply refer to the doc page in Storybook here instead of repeating it on every attribute?

components/src/web-components/visualization/gs-aggregate.stories.ts
```diff
@@ -9,6 +9,15 @@ import type { AggregateProps } from '../../preact/aggregatedData/aggregate';
 import './gs-aggregate';
 import '../app';
 
+const codeExample = `
+<gs-aggregate
```
fengelniederhammer, line 13: Do we also want to add the headline?

fengelniederhammer (changes requested):
The React example also needs an update

components/src/web-components/input/gs-mutation-filter.tsx
```diff
@@ -68,29 +68,38 @@ export class MutationFilterComponent extends PreactLitAdapter {
...
+     * The width of the component.
      *
-     * If not set, the component will take the full width of its container with height 700px.
+     * If not set, the component will take the full width of its container.
      *
-     * The width and height should be a string with a unit in css style, e.g. '100%', '500px' or '50vh'.
-     * If the unit is %, the size will be relative to the container of the component.
+     * Visit https://genspectrum.github.io/dashboards/?path=/docs/components-size-of-components--docs for more information.
      */
     @property({ type: Object })
```
fengelniederhammer, line 86:
```suggestion
    @property({ type: String })
```

components/src/web-components/visualization/gs-relative-growth-advantage.tsx
```diff
@@ -63,15 +63,24 @@ export class RelativeGrowthAdvantageComponent extends PreactLitAdapter {
...
      */
-    @property({ type: Object })
-    size: { width?: string; height?: string } | undefined = undefined;
+    @property({ type: String })
+    width: string | undefined = undefined;
+
+    /**
+     * The height of the component.
+     *
+     * If not set, the component will use the default height of 700px.
```
fengelniederhammer, line 78:
```suggestion
     * If not set, the component will use the default height of 600px.
```
At least that's what the Preact component has: `const defaultSize = { height: '600px', width: '100%' };`

components/src/web-components/visualization/gs-relative-growth-advantage.tsx
```diff
@@ -63,15 +63,24 @@ export class RelativeGrowthAdvantageComponent extends PreactLitAdapter {
...
+     *
+     * Visit https://genspectrum.github.io/dashboards/?path=/docs/components-size-of-components--docs for more information.
+     */
+    @property({ type: String })
+    height: string | undefined = undefined;
```
fengelniederhammer, line 83:
```suggestion
    width: string = '100%';

    /**
     * The height of the component.
     *
     * If not set, the component will use the default height of 700px.
     *
     * Visit https://genspectrum.github.io/dashboards/?path=/docs/components-size-of-components--docs for more information.
     */
    @property({ type: String })
    height: string = '600px';
```

fengelniederhammer: Or is there a reason not to do this? Then we could also do it for all the other components. It would simplify the docs, as we wouldn't need to mention the default size in the text.

fengelniederhammer (approved):

PR #256, opened 2024-05-21
`gs-date-range-selector`: set initial `dateFrom` and  `dateTo` values from `initialValue` 

### LINKED ISSUE #245: Date range filter: custom initial value
It should be possible to set any arbitrary date range as the initial value. (I think that's not possible at the moment?)

fengelniederhammer:
resolves #245

### PR Checklist

### DIFF
--- a/components/src/preact/dateRangeSelector/date-range-selector.tsx
+++ b/components/src/preact/dateRangeSelector/date-range-selector.tsx
@@ -2,13 +2,18 @@ import flatpickr from 'flatpickr';
 import 'flatpickr/dist/flatpickr.min.css';
 import { useEffect, useRef, useState } from 'preact/hooks';
 
+import { computeInitialValues } from './computeInitialValues';
 import { toYYYYMMDD } from './dateConversion';
+import {
+    type CustomSelectOption,
+    getDatesForSelectorValue,
+    getSelectableOptions,
+    type PresetOptionValues,
+} from './selectableOptions';
 import { ErrorBoundary } from '../components/error-boundary';
 import { ResizeContainer } from '../components/resize-container';
 import { Select } from '../components/select';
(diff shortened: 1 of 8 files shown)


JonasKellerer:
I'm not sure if this PR adresses the ticket. I think the idea was to set the inital dateFrom and initial dateTo directly. So in the end the selector is set to custom, and the values are set to the dateFrom and dateTo value. There should be no new entry as a customDateRange. @chaoran-chen ?

chaoran-chen:
Yes, exactly, @JonasKellerer! The goal is to be able to either set a special value such as "past 6 months" or a custom value such as (from 2024-01-12 to 2024-03-15).

components/src/web-components/input/gs-date-range-selector.tsx
```diff
@@ -73,6 +73,22 @@ export class DateRangeSelectorComponent extends PreactLitAdapter {
         | 'last6Months'
         | string = 'last6Months';
 
+    /**
```
JonasKellerer, line 76: Do we add a line to initialValue, that it will be overridden, when there are dateFrom or dateTo values? It is redundant but maybe useful?

components/src/preact/dateRangeSelector/date-range-selector.tsx
```diff
@@ -214,6 +227,70 @@ export const DateRangeSelectorInner = <CustomLabel extends string>({
     );
 };
 
+function computeInitialValues<CustomLabel extends string>(
```
JonasKellerer, line 230: Do we need a unit test for this?

JonasKellerer (approved):
LGTM

components/src/preact/dateRangeSelector/date-range-selector.tsx
fengelniederhammer: Done

PR #261, opened 2024-05-22
feat(components): gs-location-filter: remove submit button

### LINKED ISSUE #250: Location filter without submit button
Let's remove the submit button in the location filter component. The other input components also don't have such a button.

Instead, the component should throw an event whenever a user chooses one of the options.

fengelniederhammer:
resolves #250

### Screenshot

grafik

### PR Checklist


JonasKellerer (approved):
LGTM

Only question is, should this also fire an event, when the users just adds a blank? (It currently does, even though the value did not change).

fengelniederhammer:
I added this part to prevent sending the event when adding blanks:
```typescript
        if (inputValue.trim() === value.trim()) {
            return;
        }
```

PR #274, opened 2024-05-27
feat(components): gs-location-filter: set all values in `gs-location-changed` event details

### LINKED ISSUE #265: `gs-location-filter`: Set all fields in event
In the fired events, all fields should be present as keys (with values `undefined` if there is no value)

fengelniederhammer:
resolves  #265

### Summary

If the value is not present, explicitly set it to undefined. That way, receivers can overwrite previously set values by using the spread operator on the event details.

Event details may look like:
```typescript
{
  country: undefined,
  division: undefined,
  location: undefined,
  region: "Asia",
}
``` 

### PR Checklist


JonasKellerer (approved):
LGTM

PR #278, opened 2024-05-28
feat(components): responsive date range selector

### LINKED ISSUE #269: Less wide date range selector
It would be great to have a variant of the date range selector that is less wide. I am trying out a layout where the filters are in a left bar and it looks like this:

Could we add a two-row design where we have the special date range selection in the first and the two date input fields in the second row? It should also reduce the horizontal whitespace within each field.

~I'd suggest adding an attribute (`variant` or `layout`?) that one can set to choose between the existing and the new layout.~
Actually, should we just make it responsive? If the width is not enough, it will split the fields into two rows? We need this also when the component is shown on a small screen.

JonasKellerer:
resolves #269

### Summary

The date range selector now changes it size depending on the available size. It first shows all three elements (dropdown, date from, date to) on one row, then the dropdown on the first and the two dates on the second, and lastly all on a separate line.

### Screenshot

grafik

grafik

grafik

### PR Checklist

### DIFF
--- a/components/src/preact/dateRangeSelector/date-range-selector.tsx
+++ b/components/src/preact/dateRangeSelector/date-range-selector.tsx
@@ -182,11 +182,11 @@ export const DateRangeSelectorInner = <CustomLabel extends string>({
     };
 
     return (
-        <div class='join w-full' ref={divRef}>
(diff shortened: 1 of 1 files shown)


fengelniederhammer (approved):
LGTM

PR #282, opened 2024-05-28
263 gs location filter if value becomes empty then also fire event

### LINKED ISSUE #263: `gs-location-filter`: If value becomes empty, then also fire event
When the user deletes all input from the field, fire an event with all fields `undefined`.

JonasKellerer:
resolves #263

### Summary

location filter and text input now fire an event, when the input field is empty. The content is then the field with an "undefined" value.

### PR Checklist

### DIFF
--- a/components/src/preact/locationFilter/location-filter.tsx
+++ b/components/src/preact/locationFilter/location-filter.tsx
@@ -51,11 +51,11 @@ export const LocationFilterInner = ({ initialValue, fields }: LocationFilterInne
     const onInput = (event: JSXInternal.TargetedInputEvent<HTMLInputElement>) => {
         const inputValue = event.currentTarget.value;
         setValue(inputValue);
-        if (inputValue.trim() === value.trim()) {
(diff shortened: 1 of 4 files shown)


components/src/preact/locationFilter/location-filter.tsx
fengelniederhammer, line 1: When I type " Africa" (with a leading space), then the event with `Africa` is sent. When I delete everything up to the leading space, no event is sent. When I delete the space, still no event is sent. 

Probably we need to adapt the
```typescript
        if (inputValue.trim() === value.trim()) {
            return;
        }
``` 
part? Does `inputValue === value.trim()` do the job?

Thinking about it: `inputValue === value.trim()` also doesn't work, since typing another space after "Africa" would send another event. `inputValue.trim() === value.trim() && inputValue !== ''` maybe does it?

fengelniederhammer (approved):

components/src/preact/locationFilter/location-filter.tsx
JonasKellerer: Worked for me.

PR #295, opened 2024-06-18
feat(components): allow to set pagination limit for all tables

### LINKED ISSUE #271: For all paginated tables: make number entries per table configurable
For all paginated tables, at the moment, if the maintainer sets a height that doesn't exactly fit 10 entries, we have either too much whitespace or we have to scroll to see the pagination bar:

image

It would be great if the maintainer can define how many entries a page should have.

This applies to all paginated tables: the aggregate table component, the data tables in the different visualization components and the mutation grid.

JonasKellerer:
resolves #271

### Summary

Lets the maintainer set the paginationLimit for all tables. It can either be a number, which sets the number of rows, or a boolean. A value of  `true` produces the default pagination from gridjs. A value of `false` will disable pagination.

This targets the first step of #271. The other will be done in a separate PR.

### Screenshot

grafik

### PR Checklist

### DIFF
--- a/components/src/preact/aggregatedData/aggregate-table.tsx
+++ b/components/src/preact/aggregatedData/aggregate-table.tsx
@@ -7,9 +7,10 @@ import { formatProportion } from '../shared/table/formatProportion';
 type AggregateTableProps = {
     fields: string[];
     data: AggregateData;
+    pageSize: boolean | number;
 };
 
-export const AggregateTable: FunctionComponent<AggregateTableProps> = ({ data, fields }) => {
+export const AggregateTable: FunctionComponent<AggregateTableProps> = ({ data, fields, pageSize }) => {
     const headers = [
         ...fields.map((field) => {
             return {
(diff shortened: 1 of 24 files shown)


JonasKellerer:
There is now a follow up ticket:  #296

components/src/preact/aggregatedData/aggregate-table.tsx
```diff
@@ -28,5 +29,5 @@ export const AggregateTable: FunctionComponent<AggregateTableProps> = ({ data, f
         },
     ];
 
-    return <Table data={data} columns={headers} pagination={true} />;
+    return <Table data={data} columns={headers} paginationLimit={paginationLimit} />;
```
chaoran-chen, line 32: Suggestion for an alternative name: `pageSize`?

chaoran-chen: pageSize is shorter and sounds equally clear to me but no strong opinion :)

fengelniederhammer (approved):
LGTM

Issue #299, opened 2024-06-19
CI: Changelog Preview


fengelniederhammer:
Add a GitHub workflow that posts a changelog preview to a PR. See https://github.com/GenSpectrum/LAPIS/pull/840.

PR #305, opened 2024-06-25
refactor(components): rename (almost) all occurrences of variant to lapisFilter, or dataset

### LINKED ISSUE #253: Unify `filter`/`variant` attribute
All display components expect a `LapisFilter` attribute that will be sent to LAPIS. In some components, the attribute is called `filter`, in some `variant`. We should agree on a common name and use it in every component consistently.

JonasKellerer:
resolves #253

### Summary

Renamed all occurrences of 'variant' to 'lapisFilter', or 'dataset'. This makes the intention more concise and unifies the different namings. The only point, where variant is left over is in 'relative-growth-advantage'. There, the term variant is very closely linked to the meaning of the relative growth advantage.

### PR Checklist


fengelniederhammer (changes requested):
The code looks good, but the commit message needs to say something about a breaking change. (Let's maybe do #299 first?)

fengelniederhammer (changes requested):
The changelog looks good :+1: 

The examples need to be adapted, too.

fengelniederhammer (approved):
LGTM

PR #308, opened 2024-06-27
feat(components): show unknown date only in bar chart and table in pevalence-over-time

### LINKED ISSUE #272: Remove unknown from prevalence over time line chart
In the line chart, it doesn't make sense to have "Unknown":

image

For the other plots and for the table, it should stay.

JonasKellerer:
resolves #272

### Summary

When the dateRange is null, then unknown now shown on the table and on the bar chart. On the line and bubble chart these entries are removed.

### Screenshot

grafik

grafik

### PR Checklist


components/src/preact/prevalenceOverTime/prevalence-over-time-bar-chart.tsx
```diff
@@ -25,14 +26,24 @@ const PrevalenceOverTimeBarChart = ({
...
     const maxY =
-        yAxisScaleType !== 'logit' ? getYAxisMax(maxInData(data), yAxisMaxConfig?.[yAxisScaleType]) : undefined;
+        yAxisScaleType !== 'logit'
+            ? getYAxisMax(maxInData(nullFirstData), yAxisMaxConfig?.[yAxisScaleType])
+            : undefined;
 
     const config: ChartConfiguration = {
         type: BarWithErrorBarsController.id,
         data: {
-            labels: data[0]?.content.map((dateRange) => dateRange.dateRange?.toString() ?? 'Unknown') || [],
```
fengelniederhammer, line 34: What did those `labels` do? Apparently nothing, if we can simply delete them?

components/src/preact/prevalenceOverTime/prevalence-over-time-line-chart.tsx
```diff
@@ -100,13 +115,29 @@ const getDatasetCILower = (prevalenceOverTimeVariant: PrevalenceOverTimeVariantD
...
+    data: prevalenceOverTimeVariant.content.map((dataPoint): Datapoint => {
+        const ciLimits = wilson95PercentConfidenceInterval(dataPoint.count, dataPoint.total);
+
+        return {
+            y: dataPoint.prevalence,
+            x: dataPoint.dateRange?.toString(),
+            yCiUpper: ciLimits.upperLimit,
+            yCiLower: ciLimits.lowerLimit,
+        };
+    }),
```
fengelniederhammer, line 127: Why did this change?

fengelniederhammer (approved):

components/src/preact/prevalenceOverTime/prevalence-over-time-bar-chart.tsx
JonasKellerer: They are now put into the dataset. They are used as x labels. They must have the same length as the dataset But now the dataset contains the full datapoint. I thought this is more expressive and easier to sort together the x and y value.

components/src/preact/prevalenceOverTime/prevalence-over-time-line-chart.tsx
JonasKellerer: So we have access to the value in the tooltip, without the need to get the other dataset and find the value with the corresponding index.

PR #309, opened 2024-06-28
fix(components): correct sorting with null values in gs-aggregate

### LINKED ISSUE #270: Aggregate table: sorting is wrong
The order is wrong:

image

JonasKellerer:
resolves #270

### Summary

This fixes the issue, that the sorting with null values did not work. It seems to be a bug in gridjs.

### Screenshot

Ascending
grafik

Descending
grafik

### PR Checklist

### DIFF
--- a/components/src/preact/aggregatedData/aggregate-table.tsx
+++ b/components/src/preact/aggregatedData/aggregate-table.tsx
@@ -1,6 +1,6 @@
 import { type FunctionComponent } from 'preact';
 
-import { type AggregateData } from '../../query/queryAggregateData';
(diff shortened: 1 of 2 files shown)


fengelniederhammer (approved):
Did you file an issue at GridJS? Let's maybe also create an issue here to remove this workaround once it's fixed in GridJS?

JonasKellerer:
There is already a [gridjs issue](https://github.com/grid-js/gridjs/issues/1424), but with undefined values. I think a fix of that will resolve our problem as well. I'll add a ticket to check.

JonasKellerer:
There is now a new issue: https://github.com/GenSpectrum/dashboard-components/issues/312

PR #332, opened 2024-07-12
Number sequences over time bar chart

### LINKED ISSUE #316: Number sequences over time: bar chart

fengelniederhammer:
resolves #316

### Summary

Add the bar chart. We'll add the logarithmic scale in #333 

### Screenshot

image

image

### PR Checklist


components/tests/visualizationStories.ts
```diff
@@ -30,6 +30,11 @@ export const visualizationStories = [
     },
     { id: 'visualization-mutation-comparison--venn-diagram', title: 'Mutation comparison' },
     { id: 'visualization-aggregate--table', title: 'Aggregate', testDownloadWithFilename: 'aggregate.csv' },
+    { id: 'visualization-number-sequences-over-time--one-dataset-bar-chart', title: 'Number of sequences of time' },
     { id: 'visualization-number-sequences-over-time--one-dataset-table', title: 'Number of sequences of time' },
     { id: 'visualization-number-sequences-over-time--two-datasets', title: 'Number of sequences of time' },
+    {
+        id: 'visualization-number-sequences-over-time--two-datasets-with-non-overlapping-dates',
+        title: 'Number of sequences of time',
```
JonasKellerer, line 38:
```suggestion
        title: 'Number of sequences over time',
```

components/tests/visualizationStories.ts
```diff
@@ -30,6 +30,11 @@ export const visualizationStories = [
     },
     { id: 'visualization-mutation-comparison--venn-diagram', title: 'Mutation comparison' },
     { id: 'visualization-aggregate--table', title: 'Aggregate', testDownloadWithFilename: 'aggregate.csv' },
+    { id: 'visualization-number-sequences-over-time--one-dataset-bar-chart', title: 'Number of sequences of time' },
```
JonasKellerer, line 33:
```suggestion
    { id: 'visualization-number-sequences-over-time--one-dataset-bar-chart', title: 'Number of sequences over time' },
```

JonasKellerer (approved):
LGTM.

There are currently no errorbars, and I didn't find a related ticket. We might want to add one.

fengelniederhammer:
We decided that we don't implement error bars in this diagram (yet).

chaoran-chen:
@JonasKellerer, how would you envision an error bar? :)

JonasKellerer:
The errorbar depends on what we want to display. If we want to display a "measurement" we should include one, since all measurements are bound to have uncertainties. In this case I think of statistical uncertainty, so the user knows that there are only few measurements, but due to statistical fluctuations there could also be different values. This is important to see trends in the data and not mistake some outlier for "real". I think this is what we have done in prevalence over time. However, if we only want to provide the raw data, then no errorbar is needed.

PR #334, opened 2024-07-12
Number of sequences over time: implement line chart view

### LINKED ISSUE #317: Number sequences over time: line chart

fengelniederhammer:
resolves #317

### Summary

Adding line charts.

### Screenshot

image

image

image

### PR Checklist


chaoran-chen:
I think that filling up with 0 makes sense, is there a reason why you wouldn't fill it up?

fengelniederhammer:
> I think that filling up with 0 makes sense, is there a reason why you wouldn't fill it up?

To indicate that there is not data (as opposed to data is 0). But it doesn't matter much. For the bar chart, I had to fill everything with 0s anyway for technical reasons.

chaoran-chen:
No data in this case means 0 available sequences, so I think that we should fill up with 0.

components/src/preact/numberSequencesOverTime/number-sequences-over-time-line-chart.tsx
```diff
@@ -0,0 +1,59 @@
...
+const getDatasets = (data: NumberOfSequencesDatasets) => {
+    const tableData = getNumberOfSequencesOverTimeTableData(data, 'date');
+
+    return data.map(
+        ({ displayName }, index) =>
+            ({
+                borderWidth: 1,
+                pointRadius: 0,
+                label: displayName,
+                backgroundColor: singleGraphColorRGBAById(index, 0.3),
```
JonasKellerer, line 51: Do we need the backgroundColor when we dont have a confidence interval (yet)?

components/tests/visualizationStories.ts
```diff
@@ -31,6 +31,7 @@ export const visualizationStories = [
     { id: 'visualization-mutation-comparison--venn-diagram', title: 'Mutation comparison' },
     { id: 'visualization-aggregate--table', title: 'Aggregate', testDownloadWithFilename: 'aggregate.csv' },
     { id: 'visualization-number-sequences-over-time--one-dataset-bar-chart', title: 'Number of sequences of time' },
+    { id: 'visualization-number-sequences-over-time--one-dataset-line-chart', title: 'Number of sequences of time' },
```
JonasKellerer, line 34:
```suggestion
    { id: 'visualization-number-sequences-over-time--one-dataset-line-chart', title: 'Number of sequences over time' },
```

JonasKellerer (approved):
LGTM

components/src/preact/numberSequencesOverTime/number-sequences-over-time-line-chart.tsx
fengelniederhammer: Actually yes. It shows up in the little colored box in the tooltip.

PR #342, opened 2024-07-15
feat(components): add lineage filter component

### LINKED ISSUE #324: Add basic lineage filter with sub-lineage queries
The lineage filter component for now should be very similar to a simple text filter but with the difference that it always also suggests inputs that have a `*` suffix. As the text filter, it fetches from LAPIS all existing values and suggests the value as is and then `<value>*`.

fengelniederhammer:
resolves #324

### Summary

This is mostly copied from the text input, but I kept it separate on purpose, because the implementations might change once we support more lineage systems in this component.

### Screenshot

grafik

### PR Checklist

### DIFF
--- a/components/src/preact/lineageFilter/__mockData__/aggregated.json
+++ b/components/src/preact/lineageFilter/__mockData__/aggregated.json
@@ -1,0 +1,14510 @@
+{
+    "data": [
+        {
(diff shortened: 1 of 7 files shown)


JonasKellerer (approved):
LGTM

PR #352, opened 2024-07-17
feat(components): remove `headline` from all visualization components

### LINKED ISSUE #350: Remove headlines from components
The headlines of the components are not really useful. They cannot be styled properly.

It's easier to remove them and let users create headlines themselves outside of the component. This takes complexity from us and gives full control to the user.

fengelniederhammer:
resolves #350

### Summary

(See issue)

### PR Checklist


JonasKellerer (approved):
LGTM

Is this a breaking change? When A user has set a headline, will it still work?

And we need to adapt the headline in the dashboards repo.

fengelniederhammer:
Technically, it will probably still work, since unknown attributes would be ignored, I guess. At least I had the impression from the React example. But I would still label it breaking, because the headline disappears ;)

PR #356, opened 2024-07-18
test(components): fix expected request so that mock is hit

fengelniederhammer:
### Summary

Otherwise, the data will look unexpected because it will pull the data from the real LAPIS.

### PR Checklist


components/src/preact/numberSequencesOverTime/__mockData__/twoVariantsEG.json
```diff
@@ -244,10 +244,6 @@
             "count": 416,
             "date": "2023-09-18"
         },
-        {
```
JonasKellerer, line 247: Shouldn't we keep in some null values?

JonasKellerer (approved):
LGTM

components/src/preact/numberSequencesOverTime/__mockData__/twoVariantsEG.json
fengelniederhammer: Probably makes sense - for now I just wanted to get the tests back to green.

PR #370, opened 2024-07-22
fix(components): mutations over time: limit the number of rows and columns that can be displayed in the grid

### LINKED ISSUE #367: Mutations over time: limit number of displayed entries
This is a follow up to #59. There we decided to limit the number of displayed entries. This should be done in two ways:
1. Limit the number of dates. This can be done before the mutation queries are send. This should reduce the calculation effort and make the element faster to respond.
3. Limit the number of mutations. This can only be done after filtering for the min/max proportion. This should reduce the used memory of the browser and prevent an unresponsive state.

fengelniederhammer:
resolves #367

### Summary

Throwing a `UserFacingError` so that the user can see what went wrong.

Maybe we should adapt the error display a little so that it becomes easier to see what went wrong (instead of having to click on the "show details" button)

### Screenshot

grafik

grafik

grafik

### PR Checklist


JonasKellerer (approved):
LGTM

PR #383, opened 2024-07-24
feat(components): hide proportion text on small mutations over time windows

### LINKED ISSUE #377: Mutations over time: no horizontal scrolling
At the moment, the grid takes up more than the specified width if there are many columns, leading to horizontal scrolling:

image

That's not very useful as the idea is to see the developments at first glance. Instead, the component should be able to shrink and hide the numbers if there is not enough space. CoV-Spectrum already does that:

JonasKellerer:
resolves #377

### Summary

The proportion text on the grid is hidden, when the column width is smaller than the text.
Also changed slightly the design of the tooltip, so it is less wide. This way it will not reach to far to the right on small screens with only a few columns.

### Screenshot

grafik

### PR Checklist


fengelniederhammer (approved):
LGTM

PR #397, opened 2024-07-25
feat(components): add info texts for four components

chaoran-chen:
...namely: Aggregate, Mutations, NumberSequencesOverTime, and PrevalenceOverTime

I'll document mutations over time and mutation comparison in a later PR.

### PR Checklist

### DIFF
--- a/components/src/preact/aggregatedData/aggregate.tsx
+++ b/components/src/preact/aggregatedData/aggregate.tsx
@@ -9,7 +9,7 @@ import { CsvDownloadButton } from '../components/csv-download-button';
 import { ErrorBoundary } from '../components/error-boundary';
 import { ErrorDisplay } from '../components/error-display';
 import { Fullscreen } from '../components/fullscreen';
-import Info from '../components/info';
+import Info, { InfoHeadline1, InfoParagraph } from '../components/info';
 import { LoadingDisplay } from '../components/loading-display';
 import { NoDataDisplay } from '../components/no-data-display';
 import { ResizeContainer } from '../components/resize-container';
@@ -94,18 +94,26 @@ const AggregatedDataTabs: FunctionComponent<AggregatedDataTabsProps> = ({ data, 
 
     const tabs = views.map((view) => getTab(view));
 
-    return <Tabs tabs={tabs} toolbar={<Toolbar data={data} />} />;
+    return <Tabs tabs={tabs} toolbar={<Toolbar data={data} fields={fields} />} />;
 };
 
 type ToolbarProps = {
     data: AggregateData;
+    fields: string[];
 };
 
-const Toolbar: FunctionComponent<ToolbarProps> = ({ data }) => {
+const Toolbar: FunctionComponent<ToolbarProps> = ({ data, fields }) => {
     return (
         <div class='flex flex-row'>
             <CsvDownloadButton className='mx-1 btn btn-xs' getData={() => data} filename='aggregate.csv' />
-            <Info>Info for aggregate</Info>
+            <Info>
+                <InfoHeadline1>Aggregated data</InfoHeadline1>
+                <InfoParagraph>
+                    This table shows the number and proportion of sequences stratified by the following fields:{' '}
+                    {fields.join(', ')}. The proportion is calculated with respect to the total count within the
+                    filtered dataset.
+                </InfoParagraph>
+            </Info>
             <Fullscreen />
         </div>
     );
--- a/components/src/preact/mutations/mutations.tsx
+++ b/components/src/preact/mutations/mutations.tsx
@@ -19,7 +19,7 @@ import { CsvDownloadButton } from '../components/csv-download-button';
 import { ErrorBoundary } from '../components/error-boundary';
 import { ErrorDisplay } from '../components/error-display';
 import { Fullscreen } from '../components/fullscreen';
-import Info from '../components/info';
+import Info, { InfoHeadline1, InfoHeadline2, InfoLink, InfoParagraph } from '../components/info';
 import { LoadingDisplay } from '../components/loading-display';
 import { type DisplayedMutationType, MutationTypeSelector } from '../components/mutation-type-selector';
 import { NoDataDisplay } from '../components/no-data-display';
@@ -208,8 +208,38 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
                     filename='insertions.csv'
                 />
             )}
-            <Info>Info for mutations</Info>
+            <MutationsInfo />
             <Fullscreen />
         </>
     );
 };
+
+const MutationsInfo = () => (
+    <Info>
+        <InfoHeadline1>Mutations</InfoHeadline1>
+        <InfoParagraph>
+            This shows mutations of a variant. There are three types of mutations:{' '}
+            <InfoLink href='https://www.genome.gov/genetics-glossary/Substitution'>substitutions</InfoLink>,{' '}
+            <InfoLink href='https://www.genome.gov/genetics-glossary/Deletion'>deletions</InfoLink> and{' '}
+            <InfoLink href='https://www.genome.gov/genetics-glossary/Insertion'>insertions</InfoLink>.
+        </InfoParagraph>
+        <InfoHeadline2>Proportion calculation</InfoHeadline2>
+        <InfoParagraph>
+            The proportion of a mutation is calculated by dividing the number of sequences with the mutation by the
+            total number of sequences with a non-ambiguous symbol at the position.
+        </InfoParagraph>
+        <InfoParagraph>
+            <b>Example:</b> Assume we look at nucleotide mutations at position 5 where the reference has a T and assume
+            there are 10 sequences in total:
+            <ul className='list-disc list-inside ml-2'>
+                <li>3 sequences have a C,</li>
+                <li>2 sequences have a T,</li>
+                <li>1 sequence has a G,</li>
+                <li>3 sequences have an N,</li>
+                <li>1 sequence has a Y (which means T or C),</li>
+            </ul>
+            then the proportion of the T5C mutation is 50%. The 4 sequences that have an N or Y are excluded from the
+            calculation.
+        </InfoParagraph>
+    </Info>
+);
--- a/components/src/preact/numberSequencesOverTime/number-sequences-over-time.tsx
+++ b/components/src/preact/numberSequencesOverTime/number-sequences-over-time.tsx
@@ -1,3 +1,4 @@
+import { type FunctionComponent } from 'preact';
 import { useContext, useState } from 'preact/hooks';
 
 import { getNumberOfSequencesOverTimeTableData } from './getNumberOfSequencesOverTimeTableData';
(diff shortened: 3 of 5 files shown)


components/src/preact/aggregatedData/aggregate.tsx
```diff
@@ -93,18 +93,25 @@ const AggregatedDataTabs: FunctionComponent<AggregatedDataTabsProps> = ({ data,
...
-const Toolbar: FunctionComponent<ToolbarProps> = ({ data }) => {
+const Toolbar: FunctionComponent<ToolbarProps> = ({ data, fields }) => {
     return (
         <div class='flex flex-row'>
             <CsvDownloadButton className='mx-1 btn btn-xs' getData={() => data} filename='aggregate.csv' />
-            <Info>Info for aggregate</Info>
+            <Info>
+                <InfoHeadline1>Aggregated data</InfoHeadline1>
+                <InfoParagraph>
+                    This table shows the number and proportion of sequences stratified by the following fields:{' '}
```
fengelniederhammer, line 111: Do we need to explain what the proportion is relative to?

components/src/preact/mutations/mutations.tsx
```diff
@@ -207,7 +207,30 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
...
     );
 };
+
+const MutationsInfo = () => (
+    <Info>
+        <InfoHeadline1>Mutations</InfoHeadline1>
+        <InfoParagraph>
+            This shows mutations of a variant. There are three types of mutations:{' '}
+            <InfoLink href='https://www.genome.gov/genetics-glossary/Substitution'>substitutions</InfoLink>,{' '}
+            <InfoLink href='https://www.genome.gov/genetics-glossary/Deletion'>deletions</InfoLink>, and{' '}
```
fengelniederhammer, line 221:
```suggestion
            <InfoLink href='https://www.genome.gov/genetics-glossary/Deletion'>deletions</InfoLink> and{' '}
```

components/src/preact/numberSequencesOverTime/number-sequences-over-time.tsx
```diff
@@ -150,16 +175,25 @@ const Toolbar = ({ activeTab, data, granularity, yAxisScaleType, setYAxisScaleTy
...
+const NumberSequencesOverTimeInfo: FunctionComponent<NumberSequencesOverTimeInfoProps> = ({
+    granularity,
+    smoothingWindow,
+}) => (
     <Info>
         <InfoHeadline1>Number of sequences over time</InfoHeadline1>
         <InfoParagraph>
-            <a href='https://github.com/GenSpectrum/dashboard-components/issues/315'>TODO</a>
+            This presents the number of available sequences of a variant per <b>{granularity}</b>
+            {smoothingWindow > 0 ? `, smoothed using a ${smoothingWindow}-${granularity} sliding window` : ''}.
```
fengelniederhammer, line 196:
```suggestion
            {smoothingWindow > 0 && `, smoothed using a ${smoothingWindow}-${granularity} sliding window`}.
```

components/src/preact/prevalenceOverTime/prevalence-over-time.tsx
```diff
@@ -229,16 +238,38 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
...
+    smoothingWindow,
+    views,
+}) => {
     return (
         <Info>
             <InfoHeadline1>Prevalence over time</InfoHeadline1>
-            <InfoParagraph>Prevalence over time info.</InfoParagraph>
+            <InfoParagraph>
+                This presents the proportion of a variant per <b>{granularity}</b>
+                {smoothingWindow > 0 ? `, smoothed using a ${smoothingWindow}-${granularity} sliding window` : ''}.
```
fengelniederhammer, line 262:
```suggestion
                {smoothingWindow > 0 && `, smoothed using a ${smoothingWindow}-${granularity} sliding window`}.
```

components/src/preact/prevalenceOverTime/prevalence-over-time.tsx
```diff
@@ -229,16 +238,38 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
...
+    granularity,
+    smoothingWindow,
+    views,
+}) => {
     return (
         <Info>
             <InfoHeadline1>Prevalence over time</InfoHeadline1>
-            <InfoParagraph>Prevalence over time info.</InfoParagraph>
+            <InfoParagraph>
+                This presents the proportion of a variant per <b>{granularity}</b>
```
fengelniederhammer, line 261: I think we should explain that there is a baseline filter that we compare against here.

fengelniederhammer (changes requested):
I'll make suggestions and fix the tests.

components/src/preact/aggregatedData/aggregate.tsx
fengelniederhammer, line 1: Reminder: Squash before merge

fengelniederhammer:
I tried to clarify some descriptions a little.

components/src/preact/prevalenceOverTime/prevalence-over-time.tsx
chaoran-chen: Do you have a suggestion on how to explain it? I find it hard to come up with a general phrasing (since the baseline can be defined in very different ways).

components/src/preact/aggregatedData/aggregate.tsx
chaoran-chen: Same as below: because the baseline/denominator can be defined very flexibly, I find it difficult to come up with a good phrasing. Do you have an idea?

fengelniederhammer: "The proportion is calculated with respect to the total count within the filtered dataset."

components/src/preact/prevalenceOverTime/prevalence-over-time.tsx
fengelniederhammer:
```html
            <InfoParagraph>
                This presents the proportion of one or more variants per <b>{granularity}</b>
                {smoothingWindow > 0 && `, smoothed using a ${smoothingWindow}-${granularity} sliding window`}. The
                proportion is calculated as the number of sequences of the variant(s) divided by the total number of
                sequences of a reference variant.
            </InfoParagraph>
```

components/src/preact/aggregatedData/aggregate.tsx
chaoran-chen: "the filtered dataset" is very unspecific but once we show the exact code underlying a component (as we introduced for the prevalence over time plot in #404, we can refer to it). For now, I think adding just your sentence is good enough.

components/src/preact/mutations/mutations.tsx
```diff
@@ -207,7 +207,30 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
...
+            <InfoLink href='https://www.genome.gov/genetics-glossary/Deletion'>deletions</InfoLink> and{' '}
+            <InfoLink href='https://www.genome.gov/genetics-glossary/Insertion'>insertions</InfoLink>.
+        </InfoParagraph>
+        <InfoHeadline2>Proportion calculation</InfoHeadline2>
+        <InfoParagraph>
+            The proportion of a mutation is calculated by dividing the number of sequences with the mutation by the
+            total number of sequences with a non-ambiguous symbol at the position.
+        </InfoParagraph>
+        <InfoParagraph>
+            <b>Example:</b> If we look at nucleotide mutations at position 5 where the reference has a T: if there are
```
JonasKellerer, line 230:
```suggestion
            <b>Example:</b> Assume we look at nucleotide mutations at position 5 where the reference has a T and there are
```
Two chained ifs. And I would put some of the text in bullet points like this:

Assume we look at nucleotide mutations at position 5 where the reference has a T and there are 10 sequences in total:

- 3 sequences have a C
- 2 sequences have a T
- 1 sequence has a G
- 3 sequences have an N
- 1 sequence has a Y (which means T or C)
 
then the proportion ...

JonasKellerer: I would also add a link to the ambiguous symbols.

components/src/preact/prevalenceOverTime/prevalence-over-time.tsx
```diff
@@ -229,16 +238,40 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
...
+                This presents the proportion of one or more variants per <b>{granularity}</b>
+                {smoothingWindow > 0 && `, smoothed using a ${smoothingWindow}-${granularity} sliding window`}. The
+                proportion is calculated as the number of sequences of the variant(s) divided by the total number of
+                sequences of a reference variant.
+            </InfoParagraph>
+            {views.includes('bubble') && (
+                <>
+                    <InfoHeadline2>Bubble chart</InfoHeadline2>
+                    <InfoParagraph>
+                        In the bubble chart, the size of the bubble scales with the total number of available sequences
```
JonasKellerer, line 270:
```suggestion
                        The size of the bubble scales with the total number of available sequences
```

components/src/preact/prevalenceOverTime/prevalence-over-time.tsx
JonasKellerer: I like the bit about the smoothing window. The user can look at the code from #404 to see the lapis filter. But I think the user has no knowledge of the smoothing in that case. They see only a number, which has no context.

components/src/preact/prevalenceOverTime/prevalence-over-time.tsx
```diff
@@ -229,16 +238,40 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
...
+            <InfoParagraph>
+                This presents the proportion of one or more variants per <b>{granularity}</b>
+                {smoothingWindow > 0 && `, smoothed using a ${smoothingWindow}-${granularity} sliding window`}. The
+                proportion is calculated as the number of sequences of the variant(s) divided by the total number of
+                sequences of a reference variant.
+            </InfoParagraph>
+            {views.includes('bubble') && (
+                <>
+                    <InfoHeadline2>Bubble chart</InfoHeadline2>
+                    <InfoParagraph>
```
JonasKellerer, line 269: I think we do not mention, that in the line and bubble graph we filter for null values, and we show them in the bar and table. Maybe we can add something like

"If there are sequences, which have no assigned date, they are shown in the bar and table view."

JonasKellerer (changes requested):
Nice. Thank you. The info text is very helpful.

Do we also mention how the confidence interval is calculated? This would be the case for all display components.

components/src/preact/mutations/mutations.tsx
fengelniederhammer: Which link? Is it important to have this link here? The ambiguous symbols don't show up in the table/grid, do they?

components/src/preact/prevalenceOverTime/prevalence-over-time.tsx
chaoran-chen: Yes, I would keep the bit about the smoothing window as it is. So, together, it would be:

```
This presents the proportion of one or more variants per <b>{granularity}</b>
{smoothingWindow > 0 && `, smoothed using a ${smoothingWindow}-${granularity} sliding window`}. The
proportion is calculated as the number of sequences of the variant(s) divided by the total number 
of sequences that match the denominatorFilter (see below).
```

components/src/preact/mutations/mutations.tsx
JonasKellerer: I dont know which link. From a short google search I came up with: https://www.dnabaser.com/articles/IUPAC%20ambiguity%20codes.html (Dont know if this is a good way to present this). I just thought it would be nice to have the reference here. Or a link to our doku page. 
But your right, they dont show in the grid.

components/src/preact/mutations/mutations.tsx
```diff
@@ -207,7 +207,37 @@ const Toolbar: FunctionComponent<ToolbarProps> = ({
...
+            <InfoLink href='https://www.genome.gov/genetics-glossary/Deletion'>deletions</InfoLink> and{' '}
+            <InfoLink href='https://www.genome.gov/genetics-glossary/Insertion'>insertions</InfoLink>.
+        </InfoParagraph>
+        <InfoHeadline2>Proportion calculation</InfoHeadline2>
+        <InfoParagraph>
+            The proportion of a mutation is calculated by dividing the number of sequences with the mutation by the
+            total number of sequences with a non-ambiguous symbol at the position.
+        </InfoParagraph>
+        <InfoParagraph>
+            <b>Example:</b> Assume we look at nucleotide mutations at position 5 where the reference has a T and if
```
JonasKellerer, line 230:
```suggestion
            <b>Example:</b> Assume we look at nucleotide mutations at position 5 where the reference has a T and assume
```
I think assume would be better here. If seems wrong.

JonasKellerer (approved):

PR #400, opened 2024-07-25
ci(components): provide a standalone JS bundle

chaoran-chen:
resolves #399

### Summary

This PR changes the vite release config to create a bundle that includes all dependencies.

@fengelniederhammer, @JonasKellerer, do you see some disadvantages of doing that?

PR #401, opened 2024-07-26
fix: remove empty bullet point

chaoran-chen:


fengelniederhammer (approved):
The commit message should be something like `docs(root): ...` though.

chaoran-chen:
My actual goal is to trigger a new release.. I'm trying to release #400 but the commit was a `ci`. This doesn't seem to do it either :(

fengelniederhammer:
`fix(components):` should do the job. The Commitlint is most likely complaining about the missing scope.

But let's manually fix the generated changelog then so that at least the changelog doesn't mention this PR, but the one that changes the build.

chaoran-chen:
Ah, README says: 

> If you don't mean for a feat or fix commit to show up in the components change list, then you must not edit files in components/ in the same commit.

I need to find a bug in the components folder..

PR #405, opened 2024-07-27
feat(components): add button to show a component in fullscreen

chaoran-chen:
### Summary

This adds a button to the top-right of the toolbar of the visualization components to show a component in fullscreen mode.

### PR Checklist


components/src/preact/relativeGrowthAdvantage/relative-growth-advantage.tsx
fengelniederhammer, line 1: The chart keeps it's size after exiting full screen (on firefox):

grafik

fengelniederhammer (changes requested):
Looks really good, except for the problem with the resizing of the relative growth advantage chart.

components/src/preact/relativeGrowthAdvantage/relative-growth-advantage.tsx
fengelniederhammer: @chaoran-chen could you please check whether it works in Safari now? We tried to fix the resize issue.

chaoran-chen: In Chrome (my default browser), this works well. In Safari, I just noticed that the relative growth advantage calculation doesn't work at all (also not with the existing versions). I'll create an issue but it's independent of this PR. For the other plots, the full screen button works.

fengelniederhammer (approved):

PR #412, opened 2024-07-30
fix(components): rerender number of sequences over time when input changes 

### LINKED ISSUE #395: Number sequences over time: granularity and smooting window don't work in storybook
In [storybook](https://genspectrum.github.io/dashboard-components/?path=/story/visualization-number-sequences-over-time--one-dataset-line-chart&args=smoothingWindow:12;granularity:week), changing the granularity and the smoothing window don't seem to have any effect.

JonasKellerer:
resolves #395

### Summary

Fixes the issue that the storybook does not update the component when changing the granularity or smoothing window. There were missing dependencies in the useQuery function.

### PR Checklist


fengelniederhammer (approved):
Good catch. Maybe we should remove the default `[]` value of the dependencies argument of `useQuery`...

PR #413, opened 2024-07-30
feat(components): set toolbar components to a fixed width

### LINKED ISSUE #389: Fix toolbar componenets widths
Goal: make all toolbar components have a fixed width, so that they don't cause layout changes when selecting certain values.

Some specific components:
* segment selector: show values for: 
  * `all`
  * `none`
  * names for few selected values (question: how do we make sure that the names stay within range? We don't know how long the segments names are).
  * `X selected` for more than a certain number of selected values
* Mutations types:
  * Only show `Sub.` and `Del.` in the button that opens the dropdown as abbreviations for substitution / deletion.

JonasKellerer:
resolves #389

### Summary

The input elements `MutationType`, `SegmentSelector`, `ColorScale` and `ProportionSelectorDropdown` now have a fixed width. For `MutationType` and `SegmentSelector` the text of the button now changes dependent on the selected elements

### Screenshot

`SegmentSelector`
With many selected
grafik

With few selected
grafik

`MutationType`
With all selected
grafik

With one selected
grafik

### PR Checklist


components/src/preact/components/segment-selector.stories.tsx
```diff
@@ -0,0 +1,150 @@
...
+        const canvas = within(canvasElement);
+
+        await step('Show number of active segments as label', async () => {
+            await expect(canvas.getByText('2 segments')).toBeInTheDocument();
+        });
+    },
+};
+
+export const ShortSegmentsSelected: StoryObj<SegmentSelectorProps> = {
+    args: {
```
fengelniederhammer, line 124:
```suggestion
    ...AllSegmentsSelected,
    args: {
```

components/src/preact/components/segment-selector.tsx
```diff
@@ -16,36 +16,47 @@ export type SegmentSelectorProps = {
...
     }
 
-    return (
-        <CheckboxSelector
-            items={displayedSegments}
-            label={getSegmentSelectorLabel(displayedSegments, prefix || 'Segments: ')}
-            setItems={(items) => setDisplayedSegments(items)}
-        />
-    );
+    return `${allSelectedSelected.length} ${prefix}`;
```
fengelniederhammer, line 59: This currently shows "1 segments". I don't know how important using singular there is to us.

components/src/preact/components/segment-selector.tsx
```diff
@@ -16,36 +16,47 @@ export type SegmentSelectorProps = {
...
+    displayedSegments,
+    setDisplayedSegments,
+    prefix,
+}) => {
+    if (displayedSegments.length <= 1) {
+        return null;
+    }
+
+    return (
+        <div className='w-24'>
```
fengelniederhammer, line 29: Depending on the prefix, `w-24` can be wrong. But I think we're never passing the `prefix` anyway (maybe we should though).

fengelniederhammer (approved):

components/src/preact/components/segment-selector.tsx
JonasKellerer: I would be in favour of removing the prefix, since this does no longer makes sense with the new numbering of the elements.

components/src/preact/components/segment-selector.tsx
JonasKellerer: This is covered in the same change that removes the prefix.

Issue #421, opened 2024-07-30
Mutations over time: Add delay for proportion filter


JonasKellerer:
Currently, when users change the proportion, then the filtering step is directly initiated. Find a way to decouple this, so when the user changes the proportion filter, the changed values are shown, but the filtering is only performed after a small time period. 

A part of the solution could be 
```ts
function debounce(func: (proportion: number) => void, delayInMs: number) {
    let timeoutId: number | undefined;
    return (proportion: number) => {
        if (timeoutId) {
            clearTimeout(timeoutId);
        }
        timeoutId = window.setTimeout(() => {
            func(proportion);
        }, delayInMs);
    };
}
```

PR #422, opened 2024-07-30
perf(components): speed up mutations over time proportion filter  #388

### LINKED ISSUE #388: Mutations over time: proportion filter performance

JonasKellerer:
resolves #388

### Summary

This introduces an interface of the 2dMap. On the interface we can delete rows, by deleting only the keys of the map. The underlying data is not touched. This increases the speed because we do not need to copy the data for each filter step.

The speed now seem comparable to covSpectrum. There the calculation is hidden by showing a spinner. An idea, to create a similar user experience, is to decouple the display of the mutation proportion from the filtering process. This is targeted by this ticket: #421 

### PR Checklist


components/src/utils/Map2d.ts
```diff
@@ -73,3 +92,68 @@ export class Map2d<Key1 extends object | string, Key2 extends object | string, V
...
     }
 }
+
+export class Map2dView<Key1 extends object | string, Key2 extends object | string, Value>
+    implements Map2dInterface<Key1, Key2, Value>
+{
+    readonly keysFirstAxis;
+    readonly keysSecondAxis;
+
+    readonly baseMap: Map2d<Key1, Key2, Value> = new Map2d();
```
fengelniederhammer, line 102:
```suggestion
    readonly baseMap: Map2dInterface<Key1, Key2, Value>;
```

components/src/utils/Map2d.ts
```diff
@@ -73,3 +92,68 @@ export class Map2d<Key1 extends object | string, Key2 extends object | string, V
...
+        const serializedKeyFirstAxis = this.baseMap.serializeFirstAxis(key);
+        if (!this.keysFirstAxis.has(serializedKeyFirstAxis)) {
+            return [];
+        }
+
+        const row = this.baseMap.data.get(serializedKeyFirstAxis);
+        if (row === undefined) {
+            return [];
+        }
+        return Array.from(this.keysSecondAxis.keys()).map((key) => row.get(key) ?? fillEmptyWith);
```
fengelniederhammer, line 157:
```suggestion
        return this.baseMap.getRow(key, fillEmptyWith);
```

components/src/preact/mutationsOverTime/getFilteredMutationsOverTimeData.ts
```diff
@@ -11,7 +12,7 @@ export function getFilteredMutationOverTimeData(
     displayedMutationTypes: DisplayedMutationType[],
     proportionInterval: { min: number; max: number },
 ) {
-    const filteredData = data.copy();
```
fengelniederhammer, line 14: Let's also delete the copy method?

components/src/utils/Map2d.ts
```diff
@@ -73,3 +92,68 @@ export class Map2d<Key1 extends object | string, Key2 extends object | string, V
         return copy;
     }
 }
+
+export class Map2dView<Key1 extends object | string, Key2 extends object | string, Value>
```
fengelniederhammer, line 96: At least some basic unit test would be good.

fengelniederhammer (changes requested):
The performance is indeed a lot better, but probably still not sufficient :+1:

components/src/utils/Map2d.ts
```diff
@@ -1,6 +1,25 @@
 import hash from 'object-hash';
 
-export class Map2d<Key1 extends object | string, Key2 extends object | string, Value> {
+export interface Map2dInterface<Key1, Key2, Value> {
```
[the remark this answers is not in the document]
JonasKellerer: I tried going with `Map2dDataContainer` but I'm not very happy with it.

components/src/utils/Map2d.ts
```diff
@@ -73,3 +92,68 @@ export class Map2d<Key1 extends object | string, Key2 extends object | string, V
...
+    constructor(map: Map2dInterface<Key1, Key2, Value>) {
+        this.keysFirstAxis = new Map(map.keysFirstAxis);
+        this.keysSecondAxis = new Map(map.keysSecondAxis);
+
+        if (map instanceof Map2d) {
+            this.baseMap = map;
+        }
+        if (map instanceof Map2dView) {
+            this.baseMap = map.baseMap;
+        }
```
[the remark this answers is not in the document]
JonasKellerer: For this to work, I needed to add the `serializeFirstAxis` and `serializeSecondAxis` to the interface, which is fine for our purpose. I also added, the set method, since we renamed the interface and a map should have a set function. For the view it throws an error.

components/src/utils/Map2d.ts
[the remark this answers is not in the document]
fengelniederhammer: `BaseMap2d`? `ConcreteMap2d`?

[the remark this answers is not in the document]
JonasKellerer: I now went for `Map2dBase`

components/src/utils/map2d.spec.ts
```diff
@@ -65,30 +65,91 @@ describe('Map2d', () => {
...
+        const map2d = new Map2dBase<string, string, number>();
         map2d.set('a', 'b', 2);
-        expect(map2d.get('a', 'b')).toBe(2);
+        map2d.set('c', 'd', 4);
+
+        expect(map2d.getRow('a', 0)).toEqual([2, 0]);
+        expect(map2d.getRow('c', 0)).toEqual([0, 4]);
+    });
+
+    it('should return an epmty array when the row does not exist', () => {
```
fengelniederhammer, line 83:
```suggestion
    it('should return an empty array when the row does not exist', () => {
```

fengelniederhammer (approved):

PR #424, opened 2024-07-31
421 mutations over time add delay for proportion filter

### LINKED ISSUE #421: Mutations over time: Add delay for proportion filter

JonasKellerer:
resolves #421 

### Summary

This introduces a fix for the proportion selector, so users can input values with a comma/dot. Also this adds a delay to the proportion filter. It will now update the `setMinProportion` and `setMaxProportion` after that delay with the latest value. This way only one filter action will be performed by the `mutations-over-time` component and the `mutations` component.

### PR Checklist

- ~~[ ] The implemented feature is covered by an appropriate test.~~ But the storybook test will now wait for the new value to be set.

### DIFF
--- a/components/src/preact/components/percent-intput.tsx
+++ b/components/src/preact/components/percent-intput.tsx
@@ -21,6 +21,10 @@ export const PercentInput: FunctionComponent<PercentInputProps> = ({ percentage,
         const input = event.target as HTMLInputElement;
         const value = Number(input.value);
 
+        if (value === internalPercentage || input.value === '') {
+            return;
+        }
+
         const inRange = percentageInRange(value);
 
         if (inRange) {
--- a/components/src/preact/components/proportion-selector.tsx
+++ b/components/src/preact/components/proportion-selector.tsx
@@ -1,4 +1,5 @@
 import { type FunctionComponent } from 'preact';
+import { useEffect, useRef, useState } from 'preact/hooks';
 
 import { MinMaxRangeSlider } from './min-max-range-slider';
 import { PercentInput } from './percent-intput';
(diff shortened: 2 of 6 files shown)


components/src/preact/components/percent-intput.tsx
```diff
@@ -21,6 +21,10 @@ export const PercentInput: FunctionComponent<PercentInputProps> = ({ percentage,
         const input = event.target as HTMLInputElement;
         const value = Number(input.value);
 
+        if (value === internalPercentage) {
+            return;
+        }
```
fengelniederhammer, line 26: I'm not sure whether this really fixes the issue.

* Given the input field has value 5
* When I delete the 5 via typing a backspace
* Then the input field should be empty

Currently, the input field sets its value to 0 when deleting all input. (Unless the input was 0 before already).

components/src/preact/components/proportion-selector.tsx
```diff
@@ -11,31 +12,69 @@ export interface ProportionSelectorProps {
...
     setMinProportion,
     setMaxProportion,
 }) => {
+    const updateIntervalInMs = 300;
     const { min: minProportion, max: maxProportion } = proportionInterval;
+
+    const [internalMinProportion, setInternalMinProportion] = useState(minProportion);
+    const [internalMaxProportion, setInternalMaxProportion] = useState(maxProportion);
+
+    const minProportionRef = useRef(minProportion);
```
fengelniederhammer, line 49: As discussed: let's try whether it also works without the ref.

fengelniederhammer (changes requested):

components/src/preact/components/proportion-selector.tsx
JonasKellerer: It seems to work. I removed it.

fengelniederhammer (approved):
LGTM

components/src/preact/components/percent-intput.tsx
JonasKellerer: This is a new error. I changed it now, so when the field is empty the value is not updated. I also added tests for this.

PR #425, opened 2024-07-31
provide standalone bundle alongside bundle that needs dependencies

### LINKED ISSUE #407: Provide standalone bundle alongside bundle that needs dependencies
#399 changed the build to publish a standalone bundle only. It already contains all our dependencies. This has two major disadvantage:
* The bundle size increases (by a factor of 4 in our case)
* When installing the components via npm, then the transitive dependencies are effectively unused, because they are already bundled. This might e.g. result in the case that users cannot update transitive dependencies to fix bug or vulnerabilities in 3rd party code.

Adapt the build process to publish both - a standalone bundle that contains all dependencies as required in #399 and a version that has dependencies externalized, so that it can properly used in the npm ecosystem.
We could simply build twice with two different vite configs. We could also build a non-minified artifact that has dependencies externalized and then build again, bundling that with it's dependencies into a minified version.

fengelniederhammer:
resolves #407

### Summary

* don't bundle dependencies into the built library that npm users will use
* bundle and minify the code into a separate js file after that, publish that too.

The standalone bundle will not be includable from js code, but it's contained in the published artifact, so it should be usable via unpkg. If we notice that users need to include the standalone bundle, we can still add that later.

### Screenshot

image

### PR Checklist


JonasKellerer (approved):
LGTM. Lets test after the release in codePen

fengelniederhammer:
> Lets test after the release in codePen

I tried it - it works.

PR #426, opened 2024-08-01
docs(components): consistently test that types of web components and Preact components match

### LINKED ISSUE #363: All components: test type of prop for preact and web component
In #358 it was found that we currently only test that the type of the web component props are of specific type. It is better to test, that the type of the prop for the preact component is the same as the type of the webcomponent. This should be checked for all components.

Old:
```
type LapisFilterMatches = Expect<Equals<typeof MutationsOverTimeComponent.prototype.lapisFilter, LapisFilter>>;
```

New:
```
type LapisFilterMatches = Expect<Equals<typeof MutationsOverTimeComponent.prototype.lapisFilter, MutationsOverTimeProps["lapisFilter"]>>;
```

fengelniederhammer:
closes #363

### PR Checklist


JonasKellerer (approved):
LGTM

PR #430, opened 2024-08-02
feat(components): mutations over time: also show total number of available sequences in tooltip

### LINKED ISSUE #386: Mutations over time: show total number in tooltip
At the moment, the tooltips show the number of sequences with a mutation in a particular time frame:

In addition, it would be great to show the total number of available samples in that time frame.

fengelniederhammer:
closes #386

### Summary

* query aggregated without `fields` for every date range to get the total count
* display the total count in the tooltip

I wasn't sure how to display the tooltip exactly. I put it in the same line with the count, because an extra line uses even more space for the tooltip, which makes positioning harder.

### Screenshot

grafik

### PR Checklist

### DIFF
--- a/components/src/preact/mutationsOverTime/getFilteredMutationsOverTime.spec.ts
+++ b/components/src/preact/mutationsOverTime/getFilteredMutationsOverTime.spec.ts
@@ -15,10 +15,12 @@ describe('getFilteredMutationOverTimeData', () => {
             data.set(new Substitution('someSegment', 'A', 'T', 123), yearMonthDay('2021-01-01'), {
                 count: 1,
                 proportion: 0.1,
+                totalCount: 10,
             });
             data.set(new Substitution('someOtherSegment', 'A', 'T', 123), yearMonthDay('2021-01-01'), {
                 count: 2,
                 proportion: 0.2,
+                totalCount: 10,
             });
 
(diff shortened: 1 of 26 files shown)


components/src/preact/mutationsOverTime/getFilteredMutationsOverTime.spec.ts
```diff
@@ -76,53 +80,98 @@ describe('getFilteredMutationOverTimeData', () => {
...
         });
 
         function getMutationOverTimeData() {
             const data = new Map2dBase<Substitution | Deletion, Temporal, MutationOverTimeMutationValue>();
-            data.set(someSubstitution, yearMonthDay('2021-01-01'), { count: 1, proportion: 0.1 });
-            data.set(someSubstitution, yearMonthDay('2021-02-02'), { count: 99, proportion: 0.99 });
+            data.set(someSubstitution, yearMonthDay('2021-01-01'), {
+                count: 1,
+                proportion: 0.1,
+
```
JonasKellerer, line 167:
```suggestion
```

JonasKellerer (approved):
LGTM. Only minor things.

I like the display. I think we dont need to change it. Maybe we add this to the ticket with the info for this component. There we can explain what we mean with total.

PR #451, opened 2024-09-07
feat(components): add react support

JonasKellerer:
### Summary

I tried to add some react support for our components. I'm not sure, if this is the right way. At least the idea no longer complains. The documentation of the components also does not show, which it does in astro. 

### PR Checklist


fengelniederhammer (approved):
Let's give it a try...

Those imports will not be available when the consuming code doesn't install React, but it's only types, so it will likely still work.
`import { type DetailedHTMLProps, type HTMLAttributes } from 'react';`

PR #489, opened 2024-11-06
feat(components): gs-date-range-selector: replace `customSelectOptions` by `dateRangeOptions`

### LINKED ISSUE #473: Remove predefined date range options from DateRangeSelector
When using `gs-date-range-selector`, it becomes apparent, that the predefined date range options are hindering. One needs to write custom logic in two places. I would suggest removing the predefined options and only have "Custom" left. This way the user has more agency over the component.

fengelniederhammer:
resolves #473

### Summary

The date range selector doesn't come with predefined date ranges anymore. Users of the component need to supply themselves.
We provide the old options as constants for convenient use.

### PR Checklist

### DIFF
--- a/components/.gitignore
+++ b/components/.gitignore
@@ -1,5 +1,6 @@
 /lib/
 /dist/
+/standalone-bundle/
 /test/
 /build/
 custom-elements.json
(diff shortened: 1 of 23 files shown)


JonasKellerer (approved):
Nice. This seems way easier to use.

PR #492, opened 2024-11-07
479 display code in info box for all visualization components

### LINKED ISSUE #479: Display code in info box for all visualization components
In #404, we introduced a "Use this component yourself" section in the info box of the prevalence-over-time component. This should be added for all other visualization components as well.

fengelniederhammer:
resolves #479

### Screenshot

grafik

### PR Checklist

### DIFF
--- a/components/src/preact/aggregatedData/aggregate.tsx
+++ b/components/src/preact/aggregatedData/aggregate.tsx
@@ -9,7 +9,7 @@ import { CsvDownloadButton } from '../components/csv-download-button';
 import { ErrorBoundary } from '../components/error-boundary';
 import { ErrorDisplay } from '../components/error-display';
 import { Fullscreen } from '../components/fullscreen';
-import Info, { InfoHeadline1, InfoParagraph } from '../components/info';
+import Info, { InfoComponentCode, InfoHeadline1, InfoParagraph } from '../components/info';
 import { LoadingDisplay } from '../components/loading-display';
 import { NoDataDisplay } from '../components/no-data-display';
 import { ResizeContainer } from '../components/resize-container';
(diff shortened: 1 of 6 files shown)


components/src/preact/aggregatedData/aggregate.tsx
```diff
@@ -71,50 +63,66 @@ export const AggregateInner: FunctionComponent<AggregateInnerProps> = ({
...
+    const lapis = useContext(LapisUrlContext);
+    return (
+        <Info>
+            <InfoHeadline1>Aggregated data</InfoHeadline1>
+            <InfoParagraph>
+                This table shows the number and proportion of sequences stratified by the following fields:{' '}
+                {originalComponentProps.fields.join(', ')}. The proportion is calculated with respect to the total count
+                within the filtered dataset.
+            </InfoParagraph>
+            <InfoComponentCode componentName='aggregate' params={originalComponentProps} lapisUrl={lapis} />
```
JonasKellerer, line 125: I'm not sure when it was introduced, but the code display has current the text 'This component was created using the following parameters:` attached to it. This make no real sense to me. We should either show the parameters, or only show the code. I would favor for now the latter and just put the text below the code example above it (And leave the codepen link below)

grafik

JonasKellerer (approved):
LGTM

components/src/preact/aggregatedData/aggregate.tsx
fengelniederhammer: I actually think that it's useful. This way, it's not just any code example, it's an example to reproduce exactly what you see. By copy-pasting the example to a different page, you will have the same graph.

Issue #481, opened 2024-10-31
Allow comma separated list as input for mutation filter


JonasKellerer:
It would be nice, that users can also input a comma separated list. When users then click the "add" button, only valid values should be used, invalid values should remain in the input field (similar to cov-spectrum).

Issue #482, opened 2024-10-31
Show select dropdown on mutation filter


JonasKellerer:
When users input mutations, a dropdown (or something similar) should be shown, similar to cov-spectrum search. Users should then be able to click on that element to "log in" the mutation. This should then replace the "+" button at the side.

PR #500, opened 2024-11-13
feat(components): Use a dropdown for selecting valid mutations and allow a comma separated list as input for mutation filter

anna-parker:
resolves https://github.com/GenSpectrum/dashboard-components/issues/481, https://github.com/GenSpectrum/dashboard-components/issues/482

### Summary

1. Instead of having to press the `+` button after typing in a mutation we parse the user's input and check if it is a valid mutation. If the input is valid we show that option in a dropdown. It can be selected from the dropdown (via a click or hitting the enter button), adding it to the list of selected mutation filters. This is more intuitive for users. 

Covspectrum uses the same approach. 

The code is similar to `asyncSelect` from [react-select](https://react-select.com/home), but has been implemented in preact and is specific for the mutation filter case.

2. Allow users to paste a ',' or ';' separated list of mutations into the search bar - accept all valid mutations and leave the remaining invalid mutations on the input line.

### Screenshot

image
image

### PR Checklist

### DIFF
--- a/components/src/preact/mutationFilter/mutation-filter.stories.tsx
+++ b/components/src/preact/mutationFilter/mutation-filter.stories.tsx
@@ -1,4 +1,4 @@
-import { type Meta, type PreactRenderer, type StoryObj } from '@storybook/preact';
+import { type PreactRenderer, type Meta, type StoryObj } from '@storybook/preact';
 import { expect, fireEvent, fn, userEvent, waitFor, within } from '@storybook/test';
 import { type StepFunction } from '@storybook/types';
 
@@ -43,13 +43,75 @@ export const Default: StoryObj<MutationFilterProps> = {
     },
 };
 
+export const FiresFilterMultipleCommaSeparatedQueries: StoryObj<MutationFilterProps> = {
+    ...Default,
+    play: async ({ canvasElement, step }) => {
+        const { canvas, changedListenerMock } = await prepare(canvasElement, step);
+
+        await step('Enter a valid mutation', async () => {
+            await submitMutation(canvas, 'A123T');
+
+            await waitFor(() =>
+                expect(changedListenerMock).toHaveBeenCalledWith(
+                    expect.objectContaining({
+                        detail: {
+                            nucleotideMutations: ['A123T'],
+                            aminoAcidMutations: [],
+                            nucleotideInsertions: [],
+                            aminoAcidInsertions: [],
+                        },
+                    }),
+                ),
+            );
+        });
+
+        await step('Enter a comma separated list of valid and invalid mutations', async () => {
+            await pasteMutations(canvas, 'A123T, error_insX, A234T, ins_123:AA');
+
+            await waitFor(() =>
+                expect(changedListenerMock).toHaveBeenCalledWith(
+                    expect.objectContaining({
+                        detail: {
+                            nucleotideMutations: ['A123T', 'A234T'],
+                            aminoAcidMutations: [],
+                            nucleotideInsertions: ['ins_123:AA'],
+                            aminoAcidInsertions: [],
+                        },
+                    }),
+                ),
+            );
+            await expect(canvas.queryByText('A123T')).toBeVisible();
+            await expect(canvas.queryByText('A234T')).toBeVisible();
+            await expect(inputField(canvas)).toHaveValue('error_insX');
+        });
+
+        await step('Remove the first mutation', async () => {
+            const mutationItem = within(canvas.getByText('A234T'));
+            await fireEvent.click(mutationItem.getByRole('button', { name: '×' }));
+
+            await waitFor(() =>
+                expect(changedListenerMock).toHaveBeenCalledWith(
+                    expect.objectContaining({
+                        detail: {
+                            nucleotideMutations: ['A123T'],
+                            aminoAcidMutations: [],
+                            nucleotideInsertions: ['ins_123:AA'],
+                            aminoAcidInsertions: [],
+                        },
+                    }),
+                ),
+            );
+        });
+    },
+};
+
 export const FiresFilterChangedEvents: StoryObj<MutationFilterProps> = {
     ...Default,
     play: async ({ canvasElement, step }) => {
         const { canvas, changedListenerMock } = await prepare(canvasElement, step);
 
         await step('Enters an invalid mutation', async () => {
-            await submitMutation(canvas, 'notAMutation');
+            await testNoOptionsExist(canvas, 'notAMutation');
             await expect(changedListenerMock).not.toHaveBeenCalled();
 
             await userEvent.type(inputField(canvas), '{backspace>12/}');
@@ -88,7 +150,7 @@ export const FiresFilterChangedEvents: StoryObj<MutationFilterProps> = {
         });
 
         await step('Enter another valid mutation', async () => {
-            await submitMutation(canvas, 'ins_123:AA');
+            await submitMutation(canvas, 'ins_123:AA', 'enter');
 
             await expect(changedListenerMock).toHaveBeenCalledWith(
                 expect.objectContaining({
@@ -103,13 +165,13 @@ export const FiresFilterChangedEvents: StoryObj<MutationFilterProps> = {
         });
 
         await step('Remove the first mutation', async () => {
-            const firstMutationDeleteButton = canvas.getAllByRole('button', { name: '✕' })[1];
-            await waitFor(() => fireEvent.click(firstMutationDeleteButton));
+            const mutationItem = within(canvas.getByText('A234-'));
+            await fireEvent.click(mutationItem.getByRole('button', { name: '×' }));
 
             await expect(changedListenerMock).toHaveBeenCalledWith(
                 expect.objectContaining({
                     detail: {
-                        nucleotideMutations: ['A234-'],
+                        nucleotideMutations: ['A123T'],
                         aminoAcidMutations: [],
                         nucleotideInsertions: ['ins_123:AA'],
                         aminoAcidInsertions: [],
@@ -140,7 +202,7 @@ export const WithInitialValue: StoryObj<MutationFilterProps> = {
     play: async ({ canvasElement, step }) => {
         const { canvas, changedListenerMock } = await prepare(canvasElement, step);
 
-        await step('Enter additional input', async () => {
+        await step('Add input to initial value', async () => {
             await submitMutation(canvas, 'G500T');
 
             await expect(changedListenerMock).toHaveBeenCalledWith(
@@ -174,12 +236,34 @@ async function prepare(canvasElement: HTMLElement, step: StepFunction<PreactRend
     return { canvas, changedListenerMock };
 }
 
-const submitMutation = async (canvas: ReturnType<typeof within>, mutation: string) => {
+export type SubmissionMethod = 'click' | 'enter';
+
+const submitMutation = async (
+    canvas: ReturnType<typeof within>,
+    mutation: string,
+    submissionMethod: SubmissionMethod = 'click',
+) => {
     await userEvent.type(inputField(canvas), mutation);
-    await waitFor(() => submitButton(canvas).click());
+    const firstOption = await canvas.findByRole('option', { name: mutation });
+    if (submissionMethod === 'click') {
+        await userEvent.click(firstOption);
+    }
+    if (submissionMethod === 'enter') {
+        await userEvent.keyboard('{enter}');
+    }
+};
+
+const pasteMutations = async (canvas: ReturnType<typeof within>, mutation: string) => {
+    await userEvent.click(inputField(canvas));
+    await userEvent.paste(mutation);
+};
+
+const testNoOptionsExist = async (canvas: ReturnType<typeof within>, mutation: string) => {
+    await userEvent.type(inputField(canvas), mutation);
+    const options = canvas.queryAllByRole('option');
+
+    await expect(options).toHaveLength(0);
 };
 
 const inputField = (canvas: ReturnType<typeof within>) =>
     canvas.getByPlaceholderText('Enter a mutation', { exact: false });
-
-const submitButton = (canvas: ReturnType<typeof within>) => canvas.getByRole('button', { name: '+' });
--- a/components/src/preact/mutationFilter/mutation-filter.tsx
+++ b/components/src/preact/mutationFilter/mutation-filter.tsx
@@ -1,15 +1,10 @@
 import { type FunctionComponent } from 'preact';
-import { useContext, useRef, useState } from 'preact/hooks';
+import { useContext, useState, useRef } from 'preact/hooks';
 
 import { MutationFilterInfo } from './mutation-filter-info';
-import { parseAndValidateMutation } from './parseAndValidateMutation';
+import { parseAndValidateMutation, type ParsedMutationFilter } from './parseAndValidateMutation';
 import { type ReferenceGenome } from '../../lapisApi/ReferenceGenome';
-import {
-    type DeletionClass,
-    type InsertionClass,
-    type MutationClass,
-    type SubstitutionClass,
-} from '../../utils/mutations';
+import { type DeletionClass, type InsertionClass, type SubstitutionClass } from '../../utils/mutations';
 import { ReferenceGenomeContext } from '../ReferenceGenomeContext';
 import { ErrorBoundary } from '../components/error-boundary';
 import { singleGraphColorRGBByName } from '../shared/charts/colors';
@@ -48,37 +43,23 @@ export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> = 
     const [selectedFilters, setSelectedFilters] = useState<SelectedFilters>(
         getInitialState(initialValue, referenceGenome),
     );
-    const [inputValue, setInputValue] = useState('');
-    const [isError, setIsError] = useState(false);
-    const formRef = useRef<HTMLFormElement>(null);
 
-    const handleSubmit = (event: Event) => {
-        event.preventDefault();
-        if (inputValue === '') {
-            return;
-        }
+    const filterRef = useRef<HTMLDivElement>(null);
 
-        const parsedMutation = parseAndValidateMutation(inputValue, referenceGenome);
-
-        if (parsedMutation === null) {
-            setIsError(true);
-            return;
-        }
-
-        const newSelectedValues = {
+    const handleRemoveValue = (option: ParsedMutationFilter) => {
+        const newSelectedFilters = {
             ...selectedFilters,
-            [parsedMutation.type]: [...selectedFilters[parsedMutation.type], parsedMutation.value],
+            [option.type]: selectedFilters[option.type].filter((i) => option.value.toString() != i.toString()),
         };
 
-        setSelectedFilters(newSelectedValues);
-        fireChangeEvent(newSelectedValues);
-        setInputValue('');
+        setSelectedFilters(newSelectedFilters);
+        fireChangeEvent(newSelectedFilters);
     };
 
     const fireChangeEvent = (selectedFilters: SelectedFilters) => {
         const detail = mapToMutationFilterStrings(selectedFilters);
 
-        formRef.current?.dispatchEvent(
+        filterRef.current?.dispatchEvent(
             new CustomEvent<SelectedMutationFilterStrings>('gs-mutation-filter-changed', {
                 detail,
                 bubbles: true,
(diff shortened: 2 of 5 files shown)


fengelniederhammer (changes requested):
The behavior is slightly weird: When I type "a,2", then 2 gets accepted without any popup and "a" stays in the input field. I don't think it's terribly bad, it's just a bit weird.

Could you please clean up your code a bit? There are still some unnecessary steps from inlining the asny select component.
Also `MutationFilterInner` is too big for my taste (almost 200 lines of code). Maybe we can extract something to make it easier to digest.

And please rewrite your commit messages to something that is meaningful to users:
* Some of your commits seem to be internal refactoring, they should not be in the changelog.
* Entries in the changelog should contain references to the issue that they solve. You can add `resolves #xyz` to the commit message footer.

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -179,154 +341,6 @@ function getPlaceholder(referenceGenome: ReferenceGenome) {
     return `Enter a mutation (e.g. ${segmentPrefix}A123T, ins_${segmentPrefix}123:AT, ${firstGene}:M123E, ins_${firstGene}:123:ME)`;
 }
 
-const SelectedMutationDisplay: FunctionComponent<{
```
fengelniederhammer, line 182: Why don't we actually keep the old implementation here? Was there something wrong with it?

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -87,51 +111,189 @@ export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> =
...
+                                key={option.value.toString()}
+                                role='option'
+                                className='hover:bg-gray-300 cursor-pointer p-2 border-1 border-slate-500'
+                                onClick={() => handleOptionClick()}
+                            >
+                                {option.value.toString()}
+                            </li>
+                        )}
+                    </ul>
+                )}
```
fengelniederhammer, line 226: Thinking about it: It probably makes sense to split out this part (the `<input>` and it's attached "option" into a separate subcomponent. That would disentangle the code and encapsulate the "option" feature, so that this top-level component doesn't need to know about it.

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -87,51 +111,189 @@ export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> =
...
+            }
+
+            const newSelectedValues = {
+                ...selectedFilters,
+                [parsedMutation.type]: [...selectedFilters[parsedMutation.type], parsedMutation.value],
+            };
+            setSelectedFilters(newSelectedValues);
+            fireChangeEvent(newSelectedValues);
+            setInputValue('');
+        } else if (change.action === 'remove-value' || change.action === 'pop-value') {
```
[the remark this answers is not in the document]
anna-parker: No - I will remove it :-)

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -43,42 +38,71 @@ export const MutationFilter: FunctionComponent<MutationFilterProps> = ({ initial
...
+        setMenuVisible(false);
+    };
+
+    const handleRemoveValue = (option: ParsedMutationFilter) => {
+        handleSelectChange({ action: 'remove-value', removedValue: option });
+    };
+
+    const handleBlur = (event: FocusEvent) => {
+        // Check if click was outside the wrapper
+        if (!filterRef.current?.contains(event.relatedTarget as Node)) {
```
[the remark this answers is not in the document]
anna-parker: Otherwise blur will fire every time we click on an option as handlBlur is currently always called when we exit the input field.

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -43,42 +38,71 @@ export const MutationFilter: FunctionComponent<MutationFilterProps> = ({ initial
...
-        event.preventDefault();
-        if (inputValue === '') {
-            return;
-        }
+    const inputRef = useRef<HTMLInputElement>(null);
+    const filterRef = useRef<HTMLDivElement>(null);
 
-        const parsedMutation = parseAndValidateMutation(inputValue, referenceGenome);
+    const loadOption = (newValue: string) => {
+        return Promise.resolve(parseAndValidateMutation(newValue, referenceGenome));
```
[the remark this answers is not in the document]
anna-parker: I originally didn't add the Promise here, however without it the option that is shown in the dropdown is out of sync with the input - adding await in `handleInputChange` resolved the problem. As I'm not a typescipt expert maybe there is a better way to do this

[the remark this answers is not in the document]
fengelniederhammer: Removing the promise and the await could maybe solve it? It seems quite weird that we need to wrap sync code artificially in promises.

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -43,42 +38,71 @@ export const MutationFilter: FunctionComponent<MutationFilterProps> = ({ initial
...
     const referenceGenome = useContext(ReferenceGenomeContext);
     const [selectedFilters, setSelectedFilters] = useState<SelectedFilters>(
         getInitialState(initialValue, referenceGenome),
     );
     const [inputValue, setInputValue] = useState('');
-    const [isError, setIsError] = useState(false);
-    const formRef = useRef<HTMLFormElement>(null);
+    const [option, setOption] = useState<ParsedMutationFilter | null>(null);
+    const [loading, setLoading] = useState(false);
+    const [menuVisible, setMenuVisible] = useState(false);
```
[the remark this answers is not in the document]
anna-parker: Ah good catch!

components/src/preact/mutationFilter/mutation-filter.tsx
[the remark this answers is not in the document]
anna-parker: Oh odd it works without... must have been another issue, thanks for finding this!

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -43,42 +43,88 @@ export const MutationFilter: FunctionComponent<MutationFilterProps> = ({ initial
...
     const [inputValue, setInputValue] = useState('');
-    const [isError, setIsError] = useState(false);
-    const formRef = useRef<HTMLFormElement>(null);
+    const [option, setOption] = useState<ParsedMutationFilter | null>(null);
 
-    const handleSubmit = (event: Event) => {
-        event.preventDefault();
-        if (inputValue === '') {
-            return;
+    const inputRef = useRef<HTMLInputElement>(null);
```
fengelniederhammer, line 65: I think we don't need this, do we?

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -87,51 +133,82 @@ export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> =
...
-            <div className='absolute -top-3 -right-3'>
+        <div className='w-full border border-gray-300 rounded-md relative' ref={filterRef} onBlur={handleBlur}>
+            <div className='absolute -top-3 -right-3 z-10'>
                 <MutationFilterInfo />
             </div>
-            <div className='w-full flex p-2 flex-wrap items-center'>
-                <SelectedMutationDisplay
+
+            <div className='relative w-full' tabIndex={-1}>
+                <input
```
fengelniederhammer, line 184: I think this could use a little margin. If you look closely, it overlaps with the surrounding frame (top left corner):

grafik

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -43,42 +43,88 @@ export const MutationFilter: FunctionComponent<MutationFilterProps> = ({ initial
...
+        }
+    };
+
+    const handleBlur = (event: FocusEvent) => {
+        // Check if click was outside the wrapper
+        if (!filterRef.current?.contains(event.relatedTarget as Node)) {
+            const detail = mapToMutationFilterStrings(selectedFilters);
+
+            filterRef.current?.dispatchEvent(
+                new CustomEvent<SelectedMutationFilterStrings>('gs-mutation-filter-on-blur', {
```
fengelniederhammer, line 114: As discussed: Let's remove this event - it doesn't really make sense.

Users should be able to listen to either `-changed` events only or `-on-blur` events only. When the user deletes a mutation, no `-on-blur` event is fired (also currently on main). This is a conceptual issue -> IMO it's better to remove it completely, as it doesn't work as intended (intention: group all changes and only then fire an event).

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -43,42 +43,88 @@ export const MutationFilter: FunctionComponent<MutationFilterProps> = ({ initial
...
-    const handleSubmit = (event: Event) => {
-        event.preventDefault();
-        if (inputValue === '') {
-            return;
+    const inputRef = useRef<HTMLInputElement>(null);
+    const filterRef = useRef<HTMLDivElement>(null);
+
+    const loadOption = (newValue: string) => {
+        return parseAndValidateMutation(newValue, referenceGenome);
+    };
```
fengelniederhammer, line 70: Let's inline this function? It doesn't really provide benefit, it just obfuscates (IMO)

components/src/preact/mutationFilter/mutation-filter.tsx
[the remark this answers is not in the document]
anna-parker: Resolving as we will now remove the fire on blur dispatch event

components/src/preact/mutationFilter/mutation-filter.tsx
anna-parker: As the function is now much smaller I think I will leave this as is - but let me know if you would still like me to do this

fengelniederhammer: I would still prefer to have it split out to a separate component (for the sake of cleaner code).

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -87,38 +112,81 @@ export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> =
...
-        <form className='w-full border boder-gray-300 rounded-md relative' onSubmit={handleSubmit} ref={formRef}>
-            <div className='absolute -top-3 -right-3'>
+        <div className='w-full border border-gray-300 rounded-md relative' ref={filterRef}>
+            <div className='absolute -top-3 -right-3 z-10'>
                 <MutationFilterInfo />
             </div>
-            <div className='w-full flex p-2 flex-wrap items-center'>
-                <SelectedMutationDisplay
+
+            <div className='relative w-full p-1' tabIndex={-1}>
```
fengelniederhammer, line 162: Do we need the tabIndex?

components/src/preact/mutationFilter/mutation-filter.stories.tsx
```diff
@@ -43,13 +43,77 @@ export const Default: StoryObj<MutationFilterProps> = {
...
+                            aminoAcidInsertions: [],
+                        },
+                    }),
+                ),
+            );
+            const firstMutationButtons = canvas.queryAllByText('A123T', { exact: false });
+            expect(firstMutationButtons).toHaveLength(1);
+            const secondValidMutationButtons = canvas.queryAllByText('A234T', { exact: false });
+            expect(secondValidMutationButtons).toHaveLength(1);
+            expect(inputField(canvas)).toHaveValue('error_insX');
```
fengelniederhammer, line 87: Those expects need to be awaited:
```suggestion
            await expect(firstMutationButtons).toHaveLength(1);
            const secondValidMutationButtons = canvas.queryAllByText('A234T', { exact: false });
            await expect(secondValidMutationButtons).toHaveLength(1);
            await expect(inputField(canvas)).toHaveValue('error_insX');
```

components/src/preact/mutationFilter/mutation-filter.stories.tsx
```diff
@@ -174,12 +238,34 @@ async function prepare(canvasElement: HTMLElement, step: StepFunction<PreactRend
...
+const pasteMutations = async (canvas: ReturnType<typeof within>, mutation: string) => {
+    await userEvent.click(inputField(canvas));
+    await userEvent.paste(mutation);
+};
+
+const testNoOptionsExist = async (canvas: ReturnType<typeof within>, mutation: string) => {
+    await userEvent.type(inputField(canvas), mutation);
+    const options = canvas.queryAllByRole('option');
+
+    expect(options).toHaveLength(0);
```
fengelniederhammer, line 267:
```suggestion
    await expect(options).toHaveLength(0);
```

components/src/preact/mutationFilter/mutation-filter.stories.tsx
```diff
@@ -140,7 +204,7 @@ export const WithInitialValue: StoryObj<MutationFilterProps> = {
     play: async ({ canvasElement, step }) => {
         const { canvas, changedListenerMock } = await prepare(canvasElement, step);
 
-        await step('Enter additional input', async () => {
+        await step('Move outside of input', async () => {
```
fengelniederhammer, line 207: The step title doesn't really fit IMO.

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -43,42 +43,67 @@ export const MutationFilter: FunctionComponent<MutationFilterProps> = ({ initial
...
+const backgroundColor: { [key in keyof SelectedFilters]: string } = {
+    aminoAcidMutations: singleGraphColorRGBByName('teal', 0.4),
+    nucleotideMutations: singleGraphColorRGBByName('green', 0.4),
+    aminoAcidInsertions: singleGraphColorRGBByName('purple', 0.4),
+    nucleotideInsertions: singleGraphColorRGBByName('indigo', 0.4),
+};
+
+const backgroundColorMap = (data: ParsedMutationFilter) => {
+    return backgroundColor[data.type] || 'lightgray';
+};
```
fengelniederhammer, line 55: Let's move those two down to where they are used?

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -87,38 +112,81 @@ export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> =
         );
     };
 
-    const handleInputChange = (event: Event) => {
-        setInputValue((event.target as HTMLInputElement).value);
-        setIsError(false);
+    const handleInputChange = async (newValue: string) => {
```
fengelniederhammer, line 115:
```suggestion
    const handleInputChange = (newValue: string) => {
```

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -87,38 +112,81 @@ export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> =
...
+                <input
+                    type='text'
+                    className='w-full p-2 border-gray-300 border rounded-md'
+                    placeholder={getPlaceholder(referenceGenome)}
+                    value={inputValue}
+                    onInput={(e: Event) => {
+                        handleInputChange((e.target as HTMLInputElement).value);
+                    }}
+                    onKeyDown={(e) => handleEnterPress(e)}
+                    onFocus={() => parseAndValidateMutation(inputValue, referenceGenome)}
```
fengelniederhammer, line 172: Why do we need `onFocus` here?

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -166,150 +234,70 @@ function getPlaceholder(referenceGenome: ReferenceGenome) {
...
         <span
-            class='inline-block mx-1 px-2 py-1 font-medium text-xs rounded-full'
-            style={{ backgroundColor, color: textColor }}
+            key={mutation.toString()}
+            name={mutation.toString()}
+            className='center p-2 m-1 inline-flex text-black rounded-md'
+            style={{
+                backgroundColor: backgroundColorMap(filter),
+            }}
+            onClick={() => handleRemoveValue(filter)}
```
fengelniederhammer, line 298: I think it would be better to have a dedicated button to remove the mutation. I tried to copy a mutation, but it didn't let me do it, instead the mutation was always deleted.

fengelniederhammer (changes requested):
The current `resolves #x, #y` in the commit message doesn't work. I think you can simply make it two:
```
resolves #x
resolves #y
```

components/src/preact/mutationFilter/mutation-filter.tsx
anna-parker: Yeah this was a bit broken before, but it is needed because after blur we close the options dropdown by setting options to null. If you then click back on the input it shows no options because this is set to null, so options needs to be recalculated from the inputValue each time input comes back into focus. I modified this slightly and it works again :-)

components/src/preact/mutationFilter/mutation-filter.tsx
anna-parker: I actually need the tabIndex for making sure blur is not called just by selecting options from the dropdown - I moved it to the subfunction :-)

components/src/web-components/input/gs-mutation-filter.stories.ts
```diff
@@ -150,10 +150,14 @@ export const MultiSegmentedReferenceGenomes: StoryObj<MutationFilterProps> = {
...
+
         await waitFor(() => {
-            expect(canvas.getByText('seg1:123T')).toBeVisible();
-            expect(canvas.getByText('gene2:56')).toBeVisible();
-            return expect(canvas.getByText('ins_seg2:78:AAA')).toBeVisible();
+            const placeholderText = inputField().getAttribute('placeholder');
+
+            expect(placeholderText).toEqual(
+                'Enter a mutation (e.g. seg1:A123T, ins_seg1:123:AT, gene1:M123E, ins_gene1:123:ME)',
+            );
```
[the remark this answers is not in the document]
anna-parker: Ah I forgot about this, maybe you can actually help me with it as I cannot get this test function to work I just get shadow dom errors but the text is clearly there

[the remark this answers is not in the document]
anna-parker: Ah I fixed it now :-)

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -166,151 +245,73 @@ function getPlaceholder(referenceGenome: ReferenceGenome) {
...
-    onDelete,
-    backgroundColor,
-    textColor,
-}: SelectedFilterProps<MutationType>) => {
+const SelectedFilter = ({ handleRemoveValue, mutationFilter }: SelectedFilterProps) => {
     return (
-        <span
-            class='inline-block mx-1 px-2 py-1 font-medium text-xs rounded-full'
-            style={{ backgroundColor, color: textColor }}
+        <button
```
fengelniederhammer, line 304: In conjunction with the other change (dedicated delete button), we don't need a button here.
```suggestion
        <span
```

fengelniederhammer (changes requested):
I think we should at least do the separate delete button, otherwise it looks good :)

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -166,151 +245,73 @@ function getPlaceholder(referenceGenome: ReferenceGenome) {
...
+                backgroundColor: backgroundColorMap(mutationFilter),
+            }}
+            onClick={() => handleRemoveValue(mutationFilter)}
         >
-            {mutation.toString()}
-            <button className='ml-1' type='button' onClick={() => onDelete(mutation)}>
-                ✕
-            </button>
-        </span>
+            {mutationFilter.value.toString()} &times;
```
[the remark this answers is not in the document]
anna-parker: Good idea!

components/src/preact/mutationFilter/mutation-filter.stories.tsx
```diff
@@ -43,13 +43,77 @@ export const Default: StoryObj<MutationFilterProps> = {
...
+                            nucleotideInsertions: ['ins_123:AA'],
+                            aminoAcidInsertions: [],
+                        },
+                    }),
+                ),
+            );
+            const firstMutationButtons = canvas.queryAllByText('A123T', { exact: false });
+            await expect(firstMutationButtons).toHaveLength(1);
+            const secondValidMutationButtons = canvas.queryAllByText('A234T', { exact: false });
+            await expect(secondValidMutationButtons).toHaveLength(1);
```
fengelniederhammer, line 86:
```suggestion
            await expect(canvas.queryByText('A123T')).toBeVisible();
            await expect(canvas.queryByText('A234T')).toBeVisible();
```

fengelniederhammer (approved):
Looks good, just a few final things :)

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -158,6 +128,111 @@ function getInitialState(
...
+
+        const newSelectedValues = {
+            ...selectedFilters,
+            [option?.type]: [...selectedFilters[option?.type], option?.value],
+        };
+        setSelectedFilters(newSelectedValues);
+
+        setInputValue('');
+        setOption(null);
+    };
```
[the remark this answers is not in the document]
anna-parker: Oh no! When I initially wrote it I made sure it would deduplicate values - I must have lost that feature in one of the refactors - I will add it back

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -166,149 +241,78 @@ function getPlaceholder(referenceGenome: ReferenceGenome) {
...
+            onClick={() => handleRemoveValue(mutationFilter)}
         >
-            {mutation.toString()}
-            <button className='ml-1' type='button' onClick={() => onDelete(mutation)}>
-                ✕
+            {mutationFilter.value.toString()}
+            <button
+                className='ml-1'
+                onClick={() => handleRemoveValue(mutationFilter)}
+                name={mutationFilter.value.toString()}
```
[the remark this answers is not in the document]
anna-parker: That way you can select the buttons in the tests to delete them - I think its a good name, it should be unique when I fix the duplication bug

components/src/preact/mutationFilter/mutation-filter.stories.tsx
```diff
@@ -43,13 +43,77 @@ export const Default: StoryObj<MutationFilterProps> = {
...
+            await expect(inputField(canvas)).toHaveValue('error_insX');
+        });
+
+        await step('Remove the first mutation', async () => {
+            const firstMutationDeleteButton = canvas.getByText('A123T', { exact: false });
+            await waitFor(() => fireEvent.click(firstMutationDeleteButton));
```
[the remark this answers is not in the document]
anna-parker: Weird locally it is green for me and when I change to this it is red...

components/src/preact/mutationFilter/mutation-filter.tsx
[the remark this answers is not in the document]
anna-parker: Its fixed again :-)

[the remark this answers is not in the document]
anna-parker: Thanks for flagging this!

Issue #501, opened 2024-11-14
Mutations over time: crash when no data


fengelniederhammer:
The component fails when it has a filter that returns no data from LAPIS:
* go to mutations over time / By month story (http://localhost:6006/?path=/story/visualization-mutations-over-time--by-month)
* change lapisFilter.dateFrom to `2025-1-15`

Result:
![grafik](https://github.com/user-attachments/assets/14c38086-2842-4bdf-b34c-c0db4224fc8c)
Browser console:
![grafik](https://github.com/user-attachments/assets/9e21d195-1d6b-4be8-a24e-09c54843b728)

JonasKellerer:
#509 needs to be solved first. Moved to blocked for now.

Issue #487, opened 2024-11-06
Mutations over time: Show something when there is no data


fengelniederhammer:
Currently, the component looks like something is broken when there is no data to display:

![image](https://github.com/user-attachments/assets/d49bb92d-c228-4b6a-8376-2bea944f1f75)

When changing the proportion filter range, data appear. We should show something like "there are x mutations, but they don't match your filter".

PR #503, opened 2024-11-14
fix(components): show no data indicator on mutation over time instead of error

JonasKellerer:
Resolves: #501
Resolves: #487

### Summary
Shows a "No data available" indication, when there is no data for the lapis filters. Also adds a no data message, for the filters the user can select in the component itself.

### Screenshot

When the user selects a too narrow filter in the component
grafik

### PR Checklist

### DIFF
--- a/components/src/preact/mutationsOverTime/mutations-over-time.stories.tsx
+++ b/components/src/preact/mutationsOverTime/mutations-over-time.stories.tsx
@@ -85,3 +85,73 @@ export const ShowsMessageWhenTooManyMutations: StoryObj<MutationsOverTimeProps> 
         });
     },
 };
+
+export const ShowsNoDataWhenNoMutationsAreInFilter: StoryObj<MutationsOverTimeProps> = {
+    ...Template,
+    args: {
+        lapisFilter: { dateFrom: '1800-01-01', dateTo: '1800-01-02' },
+        sequenceType: 'nucleotide',
+        views: ['grid'],
+        width: '100%',
+        height: '700px',
+        granularity: 'year',
+        lapisDateField: 'date',
+    },
+    play: async ({ canvas }) => {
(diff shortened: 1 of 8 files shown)


fengelniederhammer (changes requested):
I'm still seeing the error and 
```
TypeError: allDates[0] is undefined
    queryOverallMutationData queryMutationsOverTime.ts:71
``` 
when I enter a `dateFrom` in 2025.

Also something seems to be weird with the webcomponents Storybook. When I change the filters there, the component doesn't really update?

JonasKellerer:
> Also something seems to be weird with the webcomponents Storybook. When I change the filters there, the component doesn't really update?

Resolved by #510

components/src/preact/mutationsOverTime/mutations-over-time.stories.tsx
```diff
@@ -85,3 +85,73 @@ export const ShowsMessageWhenTooManyMutations: StoryObj<MutationsOverTimeProps>
...
+        lapisDateField: 'date',
+    },
+    play: async ({ canvas }) => {
+        await waitFor(() => expect(canvas.getByText('No data available.', { exact: false })).toBeVisible(), {
+            timeout: 10000,
+        });
+    },
+};
+
+export const ShowsNoDataMessageWhenThereAreNoDatesInFilter: StoryObj<MutationsOverTimeProps> = {
```
fengelniederhammer, line 107: This story still calls the real LAPIS. We should mock the data to get properly stable tests.

fengelniederhammer (approved):
LGTM :+1:

components/src/preact/mutationsOverTime/mutations-over-time.stories.tsx
JonasKellerer: Good catch. I added the corresponding mock.

PR #504, opened 2024-11-14
feat(components): add mutation over time plot for wastewater / WISE

chaoran-chen:
This adds a mutation over time plot specific for the WISE wastewater data.

Preview: https://dashboard-components-git-wastewater-mutations-cov-spectrum.vercel.app/?path=/story/visualization-wastewater-mutations-over-time--default

image


components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.tsx
```diff
@@ -0,0 +1,170 @@
...
+        <Info>
+            <InfoHeadline1>Info for mutations over time</InfoHeadline1>
+            <InfoParagraph>TODO: Ask for text</InfoParagraph>
+            <InfoComponentCode componentName='mutations-over-time' params={originalComponentProps} lapisUrl={lapis} />
+        </Info>
+    );
+};
+
+function getDownloadData(data: MutationOverTimeDataPerLocation) {
+    // TODO
```
chaoran-chen, line 168: Don't forget

components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.tsx
```diff
@@ -0,0 +1,170 @@
...
+};
+
+const WastewaterMutationsOverTimeInfo: FunctionComponent<WastewaterMutationsOverTimeInfoProps> = ({
+    originalComponentProps,
+}) => {
+    const lapis = useContext(LapisUrlContext);
+    return (
+        <Info>
+            <InfoHeadline1>Info for mutations over time</InfoHeadline1>
+            <InfoParagraph>TODO: Ask for text</InfoParagraph>
```
chaoran-chen, line 161: Don't forget

components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.tsx
chaoran-chen: Can be done later.

components/src/web-components/visualization/gs-wastewater-mutations-over-time.stories.ts
```diff
@@ -0,0 +1,86 @@
...
+const codeExample = String.raw`
+<gs-wastewater-mutations-over-time
+    lapisFilter='{ "dateFrom": "2024-01-01" }'
+    sequenceType='nucleotide'
+    width='100%'
+    height='700px'
+></gs-wastewater-mutations-over-time>`;
+
+const meta: Meta<Required<WastewaterMutationsOverTimeProps>> = {
+    title: 'Visualization/Wastewater mutations over time',
```
JonasKellerer, line 20:
```suggestion
    title: 'Wastewater/Mutations over time',
```
I would put the specific wastwater components in a different folder, so it is clear, that they are not used for "normal" people.

components/src/web-components/visualization/gs-wastewater-mutations-over-time.stories.ts
JonasKellerer, line 1: When I switch the control of sequence type to "amino acid" there is no data for all locations. It would be nice to see in the story, that the entries for at least one location works.

components/src/web-components/visualization/gs-wastewater-mutations-over-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * LAPIS filter to select the displayed data.
+     */
+    @property({ type: Object })
+    lapisFilter: Record<string, string | number | null | boolean> = {};
+
+    /**
+     * Required.
+     *
+     * "nucleotide" or "amino acid"
```
JonasKellerer, line 37: This docs could tell more, but not blocking for me.

components/src/web-components/visualization/gs-wastewater-mutations-over-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+declare global {
+    // eslint-disable-next-line @typescript-eslint/no-namespace
+    namespace JSX {
+        interface IntrinsicElements {
+            'gs-wastewater-mutations-over-time': DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>;
+        }
+    }
+}
```
JonasKellerer, line 83: We don't have the type checks here, that we have for the other components.

components/src/lapisApi/lapisTypes.ts
```diff
@@ -45,6 +45,10 @@ export const aggregatedItem = z.object({ count: z.number() }).catchall(z.union([
 export const aggregatedResponse = makeLapisResponse(z.array(aggregatedItem));
 export type AggregatedItem = z.infer<typeof aggregatedItem>;
 
+export const detailsItem = z.object({}).catchall(z.union([z.string(), z.number(), z.null()]));
```
JonasKellerer, line 48: I think `z.union([z.string(), z.number(), z.null()])` is also used for the aggregatedItem. We could extract the type here, since it should be the same entries.

components/src/preact/components/tabs.tsx
```diff
@@ -16,7 +16,7 @@ const Tabs: FunctionComponent<ComponentTabsProps> = ({ tabs, toolbar }) => {
     const [activeTab, setActiveTab] = useState(tabs[0]?.title);
 
     const tabElements = (
-        <div className='flex flex-row'>
+        <div className='flex flex-row flex-wrap'>
```
JonasKellerer, line 19: I think we should change our design of the tabs a little, when we want to do this. (But this does not need to be done in this PR. When the tabs are on different rows, it would be nice to have an indicator/border around the tabs, to know that they are different tabs. Similar to the highlight on hover. 

grafik

components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+        const mutationFrequencies =
+            componentProps.sequenceType === 'nucleotide'
+                ? row.nucleotideMutationFrequency
+                : row.aminoAcidMutationFrequency;
+        for (const mutation of mutationFrequencies) {
+            map.set(
+                mutation.mutation,
+                row.date,
+                mutation.proportion !== null ? { proportion: mutation.proportion, count: NaN, totalCount: NaN } : null,
```
JonasKellerer, line 83: Why is the count and the totalCount a "NaN" should it not be a "null" or "undefined". NaN suggests to me that it is inf or something devided by zero.

components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.tsx
```diff
@@ -0,0 +1,166 @@
...
+            );
+        }
+    }
+    const mutationOverTimeDataPerLocation = [...locationMap.entries()].map(([location, data]) => ({ location, data }));
```
JonasKellerer, line 87: For me this should be in queryWasteWaterMutationsOverTime, so we can also test it.

components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.tsx
```diff
@@ -0,0 +1,166 @@
...
+    originalComponentProps: WastewaterMutationsOverTimeProps;
+};
+
+const WastewaterMutationsOverTimeInfo: FunctionComponent<WastewaterMutationsOverTimeInfoProps> = ({
+    originalComponentProps,
+}) => {
+    const lapis = useContext(LapisUrlContext);
+    return (
+        <Info>
+            <InfoHeadline1>Info for mutations over time</InfoHeadline1>
```
JonasKellerer, line 157: Not neccessary this PR, but a bit more info would be nice. Especially, since it is different from the mutations over time component.

components/src/query/queryWastewaterMutationsOverTime.ts
```diff
@@ -0,0 +1,56 @@
...
+        location: row.location as string,
+        date: toTemporalClass(parseDateStringToTemporal(row.date as string, 'day')),
+        nucleotideMutationFrequency:
+            row.nucleotideMutationFrequency !== null
+                ? transformMutations(JSON.parse(row.nucleotideMutationFrequency as string))
+                : [],
+        aminoAcidMutationFrequency:
+            row.aminoAcidMutationFrequency !== null
+                ? transformMutations(JSON.parse(row.aminoAcidMutationFrequency as string))
+                : [],
```
JonasKellerer, line 39: At some point, it would be nice to use zod here to make shure the data is correct.

JonasKellerer (changes requested):
Nice work. I am looking forward to the finished component. 

The dates in the table are not sorted. So for Lugano the second column is 17.11.2024 and the last column 16.11.2024. I think this is also the case for other locations. 

On the Laupen and Chur tab, there is only one column. In this column there are many "NoData" entries, which I find confusing. I think we should omit the rows where there is no data for all times.

components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.tsx
chaoran-chen: I would keep it as this for now and once it's up, I can point to it and ask the wastewater team whether they would like to write something to explain the method and data.

fengelniederhammer: I also created #670 to do it later.

components/src/preact/components/tabs.tsx
fengelniederhammer: I added #675

components/src/query/queryWastewaterMutationsOverTime.ts
fengelniederhammer: In the end this comes from LAPIS, which is rather trustworthy. I'm not sure whether it's worth the trouble.

components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.tsx
fengelniederhammer: I changed it to `null` to make it more explicit.

fengelniederhammer:
> On the Laupen and Chur tab, there is only one column. In this column there are many "NoData" entries, which I find confusing. I think we should omit the rows where there is no data for all times.

@chaoran-chen What do you think? Should we hide those rows? IMO it would also be ok to display them, it's a decision of what we want to show.
grafik

chaoran-chen:
I don't think that we should hide any rows. If they shouldn't be shown, the data team can just remove the data entry.

anna-parker:
As this reuses the mutations over time component when I hover over the fields I see proportion and not frequency, although the data is a frequency: 
image
can we change this?

components/src/web-components/wastewaterVisualization/gs-wastewater-mutations-over-time.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { type Equals, type Expect } from '../../utils/typeAssertions';
+import { PreactLitAdapterWithGridJsStyles } from '../PreactLitAdapterWithGridJsStyles';
+
+/**
+ * ## Context
+ *
+ * This component displays mutations for Swiss wastewater data generated within the WISE consortium. It is designed
+ * only for this purpose and is not designed to be reused outside the WISE project.
+ *
+ * It relies on a LAPIS instance that has the fields `nucleotideMutationFrequency` and `aminoAcidMutationFrequency`.
```
anna-parker, line 17: it also requires the fields `location` and `date`

fengelniederhammer: As is mentioned in the next two parts of the docs.

fengelniederhammer:
> I see proportion and not frequency, although the data is a frequency:

What do you mean by frequency?

components/src/preact/wastewater/mutationsOverTime/computeWastewaterMutationsOverTimeDataPerLocation.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+                nucleotideMutationFrequency: [
+                    { mutation: mutation1, proportion: 0.1 },
+                    { mutation: mutation2, proportion: 0.2 },
+                ],
+                aminoAcidMutationFrequency: [],
+            },
+            {
+                location: location1,
+                date: temporalCache.getYearMonthDay('2025-01-02'),
+                nucleotideMutationFrequency: [
```
anna-parker, line 32: what happens if mutation1 is not in all nucleotideMutationFrequency lists at location1? all the examples have the same nucleotideMutations in the nucleotideMutationFrequency list for the same location but this isn't mentioned in the docs as a requirement

fengelniederhammer: Missing values will result in `undefined`, which is handled in the `mutations-over-time-grid`. (It would have probably been good to add that case in a test here)

anna-parker: ah great if a mutation is not added for a specific time `null` is added in for that time and the row still exists if the mutation shows up at other time.

components/src/web-components/wastewaterVisualization/gs-wastewater-mutations-over-time.tsx
anna-parker: would be nice to have them together :-)

PR #505, opened 2024-11-14
fix(components): show no data message for prevalence over time

### LINKED ISSUE #502: Prevalence over time: crash when no data
The component crashes when LAPIS returns no data for the specified filter:
* Go to the "One Dataset" story (http://localhost:6006/?path=/story/visualization-prevalence-over-time--one-dataset)
* Change `numeratorFilter.lapisFilter.dateFrom` to `2025-10-01`

Result:
grafik
Browser console:
grafik

The error disappears when I set the smoothingWindow to 0.

JonasKellerer:
Resolves #502

### Summary
Adds a guard against missing data. It then shows "No data available". Also fixes the error, that when no data is returned the sliding window still works.

Moreover, a general guard was introduced to each of the charts (bar, line, bubble), so they don't throw an error, when they get no data.

### PR Checklist

### DIFF
--- a/components/src/preact/prevalenceOverTime/prevalence-over-time-bar-chart.tsx
+++ b/components/src/preact/prevalenceOverTime/prevalence-over-time-bar-chart.tsx
@@ -9,6 +9,7 @@ import {
 } from '../../query/queryPrevalenceOverTime';
 import { sortNullToBeginningThenByDate } from '../../utils/sort';
 import GsChart from '../components/chart';
+import { NoDataDisplay } from '../components/no-data-display';
 import { LogitScale } from '../shared/charts/LogitScale';
 import { singleGraphColorRGBAById } from '../shared/charts/colors';
 import { type ConfidenceIntervalMethod, wilson95PercentConfidenceInterval } from '../shared/charts/confideceInterval';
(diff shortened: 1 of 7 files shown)


components/src/preact/prevalenceOverTime/prevalence-over-time-bar-chart.tsx
```diff
@@ -30,12 +31,18 @@ const PrevalenceOverTimeBarChart = ({
...
+        .map((variantData) => {
+            return {
+                content: variantData.content.sort(sortNullToBeginningThenByDate),
+                displayName: variantData.displayName,
+            };
+        });
+
+    if (nullFirstData.length === 0) {
+        return <NoDataDisplay />;
+    }
```
fengelniederhammer, line 45: What data do I need to have to get into this early return? I wasn't able to find anything?

fengelniederhammer (approved):

components/src/preact/prevalenceOverTime/prevalence-over-time-bar-chart.tsx
JonasKellerer: The main point was to avoid calling `nullFirstData[0]`, when there is nothing inside. This case can occur, when the denominator filter returns an empty array. The filter step above filters for that case.

PR #507, opened 2024-11-15
fix(components): fix preact story of relative growth advantage

JonasKellerer:
### Summary

Fixes the relative growth advantage story for preact. It did not use the mock and therefore showed a different result (an error) than the web components story.

### PR Checklist


fengelniederhammer (changes requested):
LGTM - but let's change the commit message to `test` or something.

PR #511, opened 2024-11-15
feat(components): remove handleOnBlur function and dispatch event from MutationFilterInner

anna-parker:
resolves #

### Summary

In https://github.com/GenSpectrum/dashboard-components/pull/500#discussion_r1843558199 we discovered that blur dispatch events are not working as intended. (e.g. it did not fire if a user moved outside of the component if they moved outside after deleting mutations). We decided that this functionality is not needed and is tedious to implement so we will just remove it. 

The dispatch event was unused by dashboards (we use the change events dispatch event instead). 

### PR Checklist

- [ ] Corresponding PR in dashboards will need to be merged:https://github.com/GenSpectrum/dashboards/pull/301

### DIFF
--- a/components/src/preact/mutationFilter/mutation-filter.stories.tsx
+++ b/components/src/preact/mutationFilter/mutation-filter.stories.tsx
@@ -14,7 +14,7 @@ const meta: Meta<MutationFilterProps> = {
     component: MutationFilter,
     parameters: {
         actions: {
-            handles: ['gs-mutation-filter-changed', 'gs-mutation-filter-on-blur', ...previewHandles],
+            handles: ['gs-mutation-filter-changed', ...previewHandles],
         },
         fetchMock: {},
     },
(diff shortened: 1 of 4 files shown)


fengelniederhammer (changes requested):
We also need to adapt the docs accordingly (in `gs-mutation-filter.tsx`).

The commit message should indicate that this is a breaking change (be adding `BREAKING CHANGE: <explanation>` in the commit message footer)

fengelniederhammer:
> ### ⚠ BREAKING CHANGES
> 
>     * **components:** the handle was unused and inaccurate (did not fire after removing mutation queries)

This looks weird by itself - I either repeat the commit message or give instructions what to do. In this case, I would go for the instruction and say something like "removed "..." event from <component>, use ... instead".

fengelniederhammer (approved):
Please merge this with "Squash and Merge" - then we get the PR number in the changelog, since we don't have an issue for this.

PR #513, opened 2024-11-15
add reload button if something goes wrong

### LINKED ISSUE #497: Add "reload" button if something goes wrong
If an error occurs components should display a "reload" or "retry" button. This is actually meant to handle the cases when LAPIS fails, but works on a retry, because it caches the values from the previous try. For simplicity, we always offer that button.

fengelniederhammer:
resolves #497

### Screenshot

For input components:
grafik

For visualization components:
grafik

### PR Checklist


JonasKellerer (approved):
LGTM

PR #523, opened 2024-11-18
fix(components): gs-mutations-over-time: also show proportion number when switching tabs

### LINKED ISSUE #517: Mutation over time grid doesn't show numbers when switching tabs
In my PR for the wastewater mutations over time plot (#504), I'm reusing the existing mutations over time grid and am observing a bug where the proportion numbers don't display initially when switching the tab but become visible if we change the size of the window:

https://dashboard-components-git-wastewater-mutations-cov-spectrum.vercel.app/?path=/story/visualization-wastewater-mutations-over-time--default

https://github.com/user-attachments/assets/2be8b9d0-6119-4673-b9b4-f27a0d6d7149

fengelniederhammer:
resolves #517

### Summary

Making use of @container rules
https://developer.mozilla.org/en-US/docs/Web/CSS/@container

I.e. we don't rely on JS anymore to determine whether to display the number or not, which circumvents the original problem: https://github.com/GenSpectrum/dashboard-components/issues/517#issuecomment-2483403127

### PR Checklist


components/tailwind.config.cjs
fengelniederhammer, line 1: I just renamed it from `.cjs` to `.mjs` and adapted the syntax accordingly...

JonasKellerer (approved):
Nice. Good, that we use the browser features now instead of js. 

Do we also want to add a ticket, so we can use this also for the tabs, and graphs?

fengelniederhammer:
> Do we also want to add a ticket, so we can use this also for the tabs, and graphs?

I'm not sure whether it works. This solution is able to detect whether a container has a certain size, but I don't know whether you can actually get the size.

PR #528, opened 2024-11-19
fix(components): propagate UserFacingErrors through the web worker

fengelniederhammer:
### Summary

Apparently, the concrete error type doesn't make it across the wire. Any error type that you send back from a web worker will be deserialized as a plain `Error`.

### PR Checklist

### DIFF
--- a/components/src/preact/webWorkers/useWebWorker.ts
+++ b/components/src/preact/webWorkers/useWebWorker.ts
@@ -1,31 +1,53 @@
 import { useEffect, useState } from 'preact/hooks';
 
+import { UserFacingError } from '../components/error-display';
+
+export type LoadingWorkerStatus = {
+    status: 'loading';
+};
+export type SuccessWorkerStatus<Response> = {
+    status: 'success';
+    data: Response;
+};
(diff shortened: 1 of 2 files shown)


components/src/preact/webWorkers/useWebWorker.ts
```diff
@@ -1,31 +1,49 @@
...
-        worker.onmessage = (
-            event: MessageEvent<{
-                status: 'loading' | 'success' | 'error';
-                data?: Response;
-                error?: Error;
-            }>,
-        ) => {
-            const { status, data, error } = event.data;
+        worker.onmessage = (event: MessageEvent<WorkerStatus<Response>>) => {
+            const data = event.data;
```
JonasKellerer, line 33: Just nitpicking: but can we name this then "eventData". "data.data" seems wrong to me :)

JonasKellerer (approved):
Nice. 

We could improve the message to the user, for the mutations over time component. There it says, that changing the LAPIS filter or the granularity might help. This needs some rephrasing, but this is not part of this PR :)

Issue #526, opened 2024-11-19
Fire event with selected date range from date range selector


JonasKellerer:
Currently, we only provide users with an event, which can be used as a lapis filter. When using the component, it became aparent, that an event, which returns the "state" of the filter would be nice. State means the selected "label", "dateFrom" and "dateTo".

PR #529, opened 2024-11-19
feat(components): fire additional event for date range option change

JonasKellerer:
BREAKING CHANGE: Change name of event 'gs-date-range-changed' to 'gs-date-range-filter-changed'

Resolves: #526

resolves #

### PR Checklist


components/src/preact/dateRangeSelector/date-range-selector.stories.tsx
```diff
@@ -134,37 +125,83 @@ export const SetCorrectInitialDateTo: StoryObj<DateRangeSelectorProps> = {
...
-        await userEvent.type(dateFrom(), '2000-01-01');
-        await userEvent.click(dateTo());
+        step('Change date to custom value', async () => {
+            await userEvent.type(dateFromPicker(canvas), '{backspace>12}');
+            await userEvent.type(dateFromPicker(canvas), '2000-01-01');
+            await userEvent.click(dateToPicker(canvas));
+
+            await waitFor(async () => {
+                expect(selectField(canvas)).toHaveValue('Custom');
+            });
```
fengelniederhammer, line 152: You either have to `await` the `expect` or you remove the curly braces (making the `expect` the return value)
```suggestion
            await waitFor(async () => 
                expect(selectField(canvas)).toHaveValue('Custom')
            );
```

components/src/preact/dateRangeSelector/date-range-selector.stories.tsx
```diff
@@ -134,37 +125,83 @@ export const SetCorrectInitialDateTo: StoryObj<DateRangeSelectorProps> = {
...
-            expect(selectField()).toHaveValue('Custom');
+            expect(selectField(canvas)).toHaveValue('Last month');
+        });
+
+        step('Change date to custom', async () => {
+            await userEvent.selectOptions(selectField(canvas), 'CustomDateRange');
+
+            await waitFor(async () => {
+                expect(selectField(canvas)).toHaveValue('CustomDateRange');
+            });
```
fengelniederhammer, line 189: (same here)

components/src/web-components/input/gs-date-range-selector.tsx
```diff
@@ -33,6 +33,13 @@ import { PreactLitAdapter } from '../PreactLitAdapter';
  * }
  *  ```
  *  will be fired.
+ *
+ * @fires {CustomEvent<{ string | {dateFrom: string, dateTo: string}}>} gs-date-range-option-changed
+ * Fired when:
+ * - The select field is changed,
+ * - A date is selected in either of the date pickers,
+ * - A date was typed into either of the date input fields, and the input field loses focus ("on blur").
+ * Contains the selected dateRangeOption or when users select custom values it contains the selected dates.
```
fengelniederhammer, line 42: I think we should add an explanation when users should use which event (this one when you control the component from JS, the other one when you want to put the event data directly into LAPIS filters)

components/src/web-components/input/gs-date-range-selector.tsx
```diff
@@ -116,7 +128,8 @@ declare global {
     }
 
     interface HTMLElementEventMap {
-        'gs-date-range-changed': CustomEvent<Record<string, string>>;
+        'gs-date-range-filter-changed': CustomEvent<Record<string, string>>;
+        'gs-date-range-option-changed': CustomEvent<string | { dateFrom: string; dateTo: string }>;
```
fengelniederhammer, line 132: Let's extract this event into a separate type and export it, so that we can use it on the consuming side?

Then we can also dispatch the event with:
```typescript
        divRef.current?.dispatchEvent(
            event satisfies <extracted event type>
        );
```

fengelniederhammer (approved):

components/src/web-components/input/gs-date-range-selector.tsx
JonasKellerer: The event can not be a type. It must be a class. I now exported the class.

PR #533, opened 2024-11-20
feat(components): export utils as extra entrypoint

JonasKellerer:
BREAKING CHANGE: DateRangeOption and dateRangeOptionPresets now in own sub package

### Summary

When importing our components utility functions (like dateRangeOptionPresets) in a node environment or on the server side of an astro application we get an error, that "HTMLElement not defined". This message occurs, because the lit element class extends HTMLElement. 
To provide users of our package also access to our utility functions, we put them into an own sub-package. They can now be imported through e.g.: `import { dateRangeOptionPresets } from '@genspectrum/dashboard-components/util';`

### PR Checklist

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -13,7 +13,7 @@ Usage with a bundler in HTML:
 ```html
 <body>
     <script>
-        import '@genspectrum/dashboard-components';
+        import '@genspectrum/dashboard-components/components';
         import '@genspectrum/dashboard-components/style.css';
     </script>
     <gs-app lapis="https://your.lapis.url"></gs-app>
@@ -41,6 +41,13 @@ We also provide a standalone version of the components that can be used without 
 
 Internally, the components use [Preact](https://preactjs.com/).
 We use [Lit](https://lit.dev/) to create web components.
+
+We have split the package into two parts:
+
+-   The components, which are web components that can be used in the browser.
+    -   They can be imported with `import '@genspectrum/dashboard-components/components';`
+-   utility functions, which can also be used in a node environment.
+    -   They can be imported with `import '@genspectrum/dashboard-components/util';`
 
 We primarily provide two kinds of components:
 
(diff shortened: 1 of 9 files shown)


components/package.json
```diff
@@ -28,9 +28,14 @@
     },
     "exports": {
         ".": {
```
fengelniederhammer, line 30: Let's rename this to `components`, too?

components/vite.release-standalone.config.ts
```diff
@@ -5,7 +5,7 @@ export default defineConfig({
         lib: {
             formats: ['es'],
             entry: {
-                'dashboard-components': 'dist/dashboard-components.js',
+                components: 'dist/components.js',
```
fengelniederhammer, line 8: If we rename that, we also need to adapt the places where this is used (in the README, in the codepen links)

fengelniederhammer (changes requested):

components/vite.release-standalone.config.ts
JonasKellerer: I reverted it. So when importing the standalone bundle, one has still to import `dashboard-components.js`

components/package.json
JonasKellerer: Good idea. This way one does not mix them up.

components/vite.release-standalone.config.ts
```diff
@@ -5,7 +5,7 @@ export default defineConfig({
         lib: {
             formats: ['es'],
             entry: {
-                'dashboard-components': 'dist/dashboard-components.js',
+                'dashboard-components': 'dist/components.js',
```
fengelniederhammer, line 8: This only contains the components then, doesn't it? It doesn't contain the util stuff? Do we need a separate entry point, if we want to provide both?

components/README.md
```diff
@@ -51,6 +51,13 @@ We primarily provide two kinds of components:
         It is the responsibility of the dashbaord maintainer to listen to those events
         and to wire the data correctly into the visualization components.
 
+We also provide utility functions for usage with our components, which can also be used in a node environment.
```
fengelniederhammer, line 54: Do we want to add a test that it can be used in a node environment? How hard would that be?

fengelniederhammer:
I think we should also adapt the breaking changes message to something like "components: the web components must now be imported from `...components.js`,  `DateRangeOption` and `dateRangeOptionPresets` must be imported from ..."

components/vite.release-standalone.config.ts
JonasKellerer: Good catch. 
As discussed: we now have a `standaloneEntypoint.ts` file, which exports everything.

components/README.md
JonasKellerer: As discussed: we will postpone this, when we have a good idea.
I have created a ticket to keep track of this: #534

PR #541, opened 2024-11-25
fix(components): gs-mutation-filter: make input fully case-insensitive

### LINKED ISSUE #536: Mutation filter should be fully case insensitive
At the moment, it seems that the mutation filter is case in sensitive w.r.t. to the bases (e.g., `e627k` is just as accepted as `E627K`) but it is case sensitive w.r.t. to the segment name (e.g., `PB2:…` is accepted but not `pb2:…` for influenza). However, it should be case insensitive.

fengelniederhammer:
resolves #536

### Screenshot

grafik

### PR Checklist

### DIFF
--- a/components/src/preact/mutationFilter/parseMutation.spec.ts
+++ b/components/src/preact/mutationFilter/parseMutation.spec.ts
@@ -36,6 +36,16 @@ describe('parseMutation', () => {
                 expected: { type: 'aminoAcidInsertions', value: new InsertionClass('gene1', 10, 'ACGT') },
             },
             {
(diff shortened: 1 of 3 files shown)


JonasKellerer (approved):
LGTM.

Do we want to show the users, what they have inserted, or everything in capial letters? I just found your example picture hard to read. But maybe, when users input it like this, they also want to be shown it like this?

fengelniederhammer:
> when users input it like this, they also want to be shown it like this?

I think that's what users want ;)

PR #542, opened 2024-11-25
feat(components): gs-prevalence-over-time: don't show confidence interval by default

### LINKED ISSUE #537: Prevalence over time: Don't show confidence interval by default
In our datasets, except for COVID where we have a lot of sequences, the confidence intervals don't really make sense. Firstly, because there are usually so few sequences, the CIs tend to fill most of the plot. Secondly, we can less assume that the data are (close to) randomly sampled. So, we shouldn't show the CIs by default.

fengelniederhammer:
resolves #537

### Screenshot

grafik

### PR Checklist


JonasKellerer (approved):
LGTM

PR #553, opened 2024-11-29
Check how we could implement a map component

### LINKED ISSUE #548: Map: sketch a solution
Think about how we can implement a map component.
* Which technical solution can we use? How do we get the graphical charts?
* how do we map names in the data to the entities on the map?
* What do we show on the map?

fengelniederhammer:
resolves #548

### Summary

What I found out about the data itself: There are two widely used data formats:
* https://geojson.org/
* https://github.com/topojson/topojson (claims to be a more compact way of storing the data)
  * They offer a JS library to convert topojson to geojson
Basically all visualization libraries that I found support geojson.

I found https://github.com/markmarkoh/datamaps/tree/master/src/js/data, which has map data for many countries (in the topojson format). The repo itself looks pretty unmaintained, but most countries don't change that often. Taking the map data should still be ok.

I've looked into https://www.npmjs.com/package/chartjs-chart-geo, but it wasn't really satisfying. I couldn't warp my head around how to properly configure the projections to make the maps look good (centered, proportions not distorted too much).

The current approach is based on https://leafletjs.com/. It seems quite powerful and very customizable. 

The main open points to me are:
* Do we provide the maps? Do we offer a possibility to let users plug in the maps?
  * The maps are quite significant in size (at least in browser terms): Usually around ~100-500 kB
* How do we map the values from LAPIS to the states in the maps? We don't have control over the LAPIS instance -> we probably need a way to let the user provide a mapping between lapis field values and state names in the maps.

### PR Checklist


anna-parker:
I've always plotted maps in python with geopandas, and in the past used shapefiles - but it does seem like switching to geoJSON for typescript would be best.

As far as where we could get maps from: from my google search it looks like https://gadm.org/download_country.html and https://www.naturalearthdata.com/downloads/ are often used for this. GADM uses official administrative region names or IDs - we could get these if we tried to map data to their official names e.g. using GeoNames: https://www.geonames.org/countries. I've looked into this quite a bit for Loculus - the issue here is that INSDC does not validation geographic information apart from country. So we will have to do some data processing/clean-up to use the data from INSDC - hopefully as we are using Loculus we can do the processing there and then use the output in Genspectrum.

fengelniederhammer:
I think this component is now mostly ready for review. Some docs and tests are still missing. Other than that, there are some larger things that we should investigate but can be postponed to later issues:
* review the color scale and add a legend
* do we want to remove the little Leaflet finterprint at the bottom right?
* Zoom and offset might need to be adapted when the size of the component changes (i.e. making the component narrow could cut off the map on the left and right)
* Maybe we should outsource some computation work to a web worker?
* Maybe it makes sense to offer map bounds to prevent scrolling too far.

components/src/preact/map/map-component-map-view.tsx
```diff
@@ -0,0 +1,179 @@
...
+        { limit: 0.4, color: '#BD0026' },
+        { limit: 0.3, color: '#E31A1C' },
+        { limit: 0.2, color: '#FC4E2A' },
+        { limit: 0.1, color: '#FD8D3C' },
+        { limit: 0.05, color: '#FEB24C' },
+        { limit: 0.02, color: '#FED976' },
+        { limit: 0.01, color: '#FFEDA0' },
+        { limit: 0.005, color: '#FFF7BC' },
+        { limit: 0.002, color: '#FFFFCC' },
+    ];
```
JonasKellerer, line 137: We could use colorblind friendly colors here. https://personal.sron.nl/~pault/#sec:sequential . They are similar to these colors.

components/src/preact/map/map.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+        lapisLocationField: 'country',
+        mapSource: {
+            type: 'topojson',
+            url: worldMapUrl,
+            topologyObjectsKey: 'countries',
+        },
+        enableMapNavigation: false,
+        width: '1100px',
+        height: '800px',
+        views: ['map'],
```
JonasKellerer, line 38:
```suggestion
        views: ['map'],
        zoom: 2,
        offsetX: 0,
        offsetY: 10,
```

components/src/preact/map/map.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+type ToolbarProps = {
+    originalComponentProps: MapProps;
+};
+
+const Toolbar: FunctionComponent<ToolbarProps> = ({ originalComponentProps }) => {
+    return (
+        <div class='flex flex-row'>
+            <MapInfo originalComponentProps={originalComponentProps} />
+            <Fullscreen />
```
JonasKellerer, line 121: The fullscreen button does not work for me in storybook. Is this a know bug?

components/src/preact/map/useGeoJsonMap.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export type GeoJsonFeatureProperties = {
+    name: string;
+};
+
+async function loadTopojsonMap(
+    mapSource: MapSource,
+): Promise<FeatureCollection<GeometryObject, GeoJsonFeatureProperties>> {
+    const response = await fetch(mapSource.url);
+    const topology = (await response.json()) as Topology;
```
JonasKellerer, line 39: Can we use zod here?

JonasKellerer (commented):
Looks great. 

To discuss:
  - name of the component (I would prefer something without the name map inside. When I recall it correctly, we also want to show a table of the data. So maybe `gs-geographic-distribution`?)
  - are there other projections of the world map? Or can be let the user decide on the projection?
 
Some other points for the future
  - show color scale
  - a zoom reset button would be nice

components/src/preact/map/useGeoJsonMap.tsx
fengelniederhammer: No - we don't have a zod type for `Topology`which comes from the topojson types. And I would like to avoid writing one ourselves.

components/src/preact/map/map.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { LoadingDisplay } from '../components/loading-display';
+import { ResizeContainer } from '../components/resize-container';
+import { useQuery } from '../useQuery';
+import { type GeoJsonFeatureProperties, useGeoJsonMap } from './useGeoJsonMap';
+import { lapisFilterSchema, type MapView, mapViewSchema, views } from '../../types';
+import Tabs from '../components/tabs';
+
+const mapSourceSchema = z.object({
+    type: z.literal('topojson'),
+    url: z.string().min(1),
```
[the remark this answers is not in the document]
fengelniederhammer: I think I would actually like to leave it as it is - `.url()` doesn't allow relative links, which will be useful for us when we host the map on the same server.

components/src/preact/map/map.tsx
fengelniederhammer: I just checked the implementation - the problem is inherent to it. It tries to find the next shadow root, which doesn't work in the Preact storybook.

JonasKellerer: I think we can leave it like it is. I just did not know if we already encounreted this issue.

components/src/preact/map/map.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+const MapTabs: FunctionComponent<MapTabsProps> = ({ originalComponentProps, geojsonData, data }) => {
+    const getTab = (view: MapView) => {
+        switch (view) {
+            case views.map:
+                return {
+                    title: 'Map',
+                    content: (
+                        <MapComponentMapView
+                            locationData={data}
```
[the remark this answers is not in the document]
fengelniederhammer: Actually it's not "by location" yet. It's just the array that LAPIS returns.

components/src/preact/map/map.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+        lapisFilter: { dateFrom: '2022-01-01', dateTo: '2022-04-01' },
+        lapisLocationField: 'country',
+        mapSource: {
+            type: 'topojson',
+            url: worldMapUrl,
+            topologyObjectsKey: 'countries',
+        },
+        enableMapNavigation: false,
+        width: '1100px',
+        height: '800px',
```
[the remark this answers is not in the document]
fengelniederhammer: Yes, and I don't know how to fix that yet. I think for now it's better to fix the width until we find a proper solution.

components/src/preact/map/map.tsx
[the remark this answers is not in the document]
JonasKellerer: I was confused, why there are two objects with location data. The data from lapis is just data stratified by location and not different locations. But we can leave it like this.

components/src/preact/map/map-component-map-view.tsx
fengelniederhammer: The scale is more or less arbitrary. I'm not sure how to best scale this. It highly depends on the data...

I added an entry to the list in #547 so that we can come back to this later.

fengelniederhammer:
> * are there other projections of the world map? Or can be let the user decide on the projection?

Leaflet allows setting the projection. But I don't know whether we want to burden a user with that.

fengelniederhammer:
All TODOs that are immediately necessary for a first working version of this component should be resolved now.

I created tasks in #547 for everything that can be done later.

components/src/preact/map/sequences-by-location.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+type SequencesByLocationMapInfoProps = {
+    originalComponentProps: SequencesByLocationProps;
+};
+
+const SequencesByLocationMapInfo: FunctionComponent<SequencesByLocationMapInfoProps> = ({ originalComponentProps }) => {
+    const lapis = useContext(LapisUrlContext);
+    return (
+        <Info>
+            <InfoHeadline1>Prevalence By Location</InfoHeadline1>
```
JonasKellerer, line 133:
```suggestion
            <InfoHeadline1>Sequence by location</InfoHeadline1>
```
So far we have only capitalized the first letter.

components/src/web-components/visualization/gs-sequences-by-location.tsx
```diff
@@ -0,0 +1,234 @@
...
+     */
+    @property({ type: Number })
+    zoom: number = 1;
+
+    /**
+     * Initially shift the center of the map in x direction (longitude).
+     *
+     * `-180` is the International Date Line with the map shifted to the right, `0` is the prime meridian,
+     * `180` is the International Date Line with the map shifted to the left.
+     */
```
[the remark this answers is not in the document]
fengelniederhammer: It probably depends on the projection. But I didn't find a way yet such that one does not need zoom and offset.

Also the maps that I used put the countries in the "right place" on the world map -> They are not centered.

components/src/web-components/visualization/gs-sequences-by-location.tsx
```diff
@@ -0,0 +1,234 @@
...
+    @property({ type: Boolean })
+    enableMapNavigation: boolean = true;
+
+    /**
+     * The width of the component.
+     *
+     * Visit https://genspectrum.github.io/dashboard-components/?path=/docs/components-size-of-components--docs for more information.
+     */
+    @property({ type: String })
+    width: string = '100%';
```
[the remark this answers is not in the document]
fengelniederhammer: It's not good. But I think a fixed pixel size is not better...

PR #558, opened 2024-12-04
fix(components): gs-date-range-selector: validate component attributes

### LINKED ISSUE #557: gs-date-range-selector: validate input

fengelniederhammer:
resolves #557

### Summary

This is a proof of concept how we can use zod to validate component attributes.

### Screenshot

grafik

### PR Checklist

### DIFF
--- a/components/src/preact/components/error-display.tsx
+++ b/components/src/preact/components/error-display.tsx
@@ -1,5 +1,6 @@
 import { type FunctionComponent } from 'preact';
 import { useEffect, useRef } from 'preact/hooks';
+import { type ZodError } from 'zod';
 
 import { LapisError, UnknownLapisError } from '../../lapisApi/lapisApi';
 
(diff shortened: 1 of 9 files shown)


components/src/preact/components/error-display.tsx
```diff
@@ -115,5 +125,21 @@ function getDisplayedErrorMessage(error: Error) {
         };
     }
 
+    if (error instanceof InvalidPropsError) {
+        const firstError = error.zodError.errors[0];
+        let message = error.zodError.issues
+            .map((issue) => `Unexpected value for "${issue.path}": ${issue.message}`)
```
JonasKellerer, line 131:
```suggestion
           .map((issue) => {
                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
                // @ts-expect-error
                return `Unexpected value "${JSON.stringify(error.componentProps[issue.path])}" for "${issue.path}": ${issue.message}`;
            })
```
It would be nice to show also the value, which is wrong. I did not find a better solution, than to use the "ts-expect-error".

components/src/preact/components/error-display.tsx
```diff
@@ -24,9 +25,16 @@ export class UserFacingError extends Error {
     }
 }
 
+export class InvalidPropsError extends Error {
+    constructor(public readonly zodError: ZodError) {
+        super(zodError.message);
+        this.name = 'InvalidPropsError';
+    }
+}
+
```
JonasKellerer, line 34:
```suggestion
export class InvalidPropsError extends Error {
    constructor(public readonly zodError: ZodError, public readonly componentProps: unknown) {
        super(zodError.message);
        this.name = 'InvalidPropsError';
    }
}
```
See comment below.

JonasKellerer (approved):
Nice. Good to have. 

Maybe we also show what is the conflicting value?

PR #567, opened 2024-12-06
565 mutation over time display missing values differently

### LINKED ISSUE #565: Mutation over time: display missing values differently
Sometimes, due to missing data, we don't know the proportion of a mutation at a certain date. In such a case, the mutations over time component should show it in a different color (e.g., gray?) instead of considering the values as 0%.

fengelniederhammer:
resolves #565

### Screenshot

@chaoran-chen Is that how you expected it to look like?
image

### PR Checklist


chaoran-chen:
Yes, this looks great! Thanks!

JonasKellerer (approved):
Looks great.
Nice change, that we now can store undefined values. This should also reduce the size of the map.

PR #590, opened 2024-12-09
fix(components): gs-prevalence-over-time: validate attributes

JonasKellerer:
Resolves: #576

### PR Checklist

### DIFF
--- a/components/src/preact/shared/charts/confideceInterval.ts
+++ b/components/src/preact/shared/charts/confideceInterval.ts
@@ -1,10 +1,11 @@
-// calculateWilsonInterval calculates the Wilson score interval for 95% confidence.
-//
-// This function is based on https://github.com/erikfox/wilson-interval, but without
-// high precision math.
-//
-// observed - number of observed positive outcomes
-// sample - number of experiments or size of the sample
+import z from 'zod';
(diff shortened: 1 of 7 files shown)


components/src/preact/shared/charts/confideceInterval.ts
```diff
@@ -5,6 +5,8 @@
 //
 // observed - number of observed positive outcomes
 // sample - number of experiments or size of the sample
+import z from 'zod';
```
fengelniederhammer, line 8: Aha, classic example of how comments get detached from the line that they are supposed to comment ;)

Let's make the comment a docstring (`/** ... */`) and hope that IDEs respect its attachment to the function in the future.

fengelniederhammer (approved):

components/src/web-components/visualization/gs-prevalence-over-time.tsx
fengelniederhammer, line 1: Let's remove the default value from `lapisDateField`

JonasKellerer: Good catch. I thought, I got the all :see_no_evil:

PR #600, opened 2024-12-10
fix(components): update error boundary to validate all props

JonasKellerer:
Resolves: #599

### PR Checklist

### DIFF
--- a/components/src/preact/components/error-boundary.tsx
+++ b/components/src/preact/components/error-boundary.tsx
@@ -7,9 +7,9 @@ import { ResizeContainer, type Size } from './resize-container';
 
 type ErrorBoundaryProps<T> = {
     size: Size;
+    componentProps: T;
+    schema: ZodSchema<T>;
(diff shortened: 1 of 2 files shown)


components/src/preact/components/error-boundary.tsx
```diff
@@ -41,7 +41,6 @@ export const ErrorBoundary = <T extends Record<string, unknown>>({
     return <>{children}</>;
 };
 
-// TODO #554 - make both arguments required once all components validate their props
 function useCheckComponentProps<T extends Record<string, unknown>>(schema?: ZodSchema<T>, componentProps?: T) {
```
fengelniederhammer, line 44:
```suggestion
function useCheckComponentProps<T extends Record<string, unknown>>(schema: ZodSchema<T>, componentProps: T) {
```

fengelniederhammer (approved):
The commit message should be `refactor`or `chore` - this is nothing that users of the library would notice.

PR #602, opened 2024-12-11
feat(components): gs-sequences-by-location: add table view

### LINKED ISSUE #597: gs-sequences-by-location: add table view

fengelniederhammer:
resolves #597

### Screenshot

image

### PR Checklist


components/src/preact/map/sequences-by-location.tsx
```diff
@@ -13,24 +12,25 @@ import Info, { InfoComponentCode, InfoHeadline1, InfoParagraph } from '../compon
...
-    mapSource: mapSourceSchema,
+    mapSource: mapSourceSchema.optional(),
     enableMapNavigation: z.boolean(),
     width: z.string(),
     height: z.string(),
     views: z.array(sequencesByLocationViewSchema),
     zoom: z.number(),
     offsetX: z.number(),
     offsetY: z.number(),
+    pageSize: z.union([z.boolean(), z.number()]),
```
JonasKellerer, line 33:
```suggestion
    pageSize: z.union([z.boolean(), z.number()]).optional(),
```
Is this also an optional value, when users dont want to show the table?

The same for the zoom, offsetX, offsetY and enableMapNavigation.

JonasKellerer: I think this is the first time i realized this, but the props of the view and the view list should be the same. Maybe we keep it like this and add a ticket to change this. I could imagine, that we put them into one prop:
```ts
...
lapisFilter: {dateFrom: .... },
lapisLocationField: "country",
views: [
  { type: "map",  mapSource: mapSourceSchema ...}
  { type: "table", pageSize: true }
]
...
```
But we would need to change this for all our other components as well. And we would need to have different props for the web component and the preact component. Or we also change this for the web component.

components/src/web-components/visualization/gs-sequences-by-location.tsx
```diff
@@ -81,10 +81,19 @@ const leafletModificationsCss = unsafeCSS(leafletStyleModifications);
...
  *   This is expected, as LAPIS will only return locations where sequences have been collected for that filter.
+ *
+ * ### Table View
+ *
+ * This view displays the data in a table format.
+ * It is similar to the table view of the `gs-aggregate` component.
+ * The table has three columns:
+ * - `lapisLocationField`,
+ * - `count` (the number of samples in this location matching the `lapisFilter`),
+ * - `proportion` (`count` / sum of the `count` of all locations).
```
JonasKellerer, line 92: Do we want to add, that only in this view, users can see the null values?

JonasKellerer (approved):
LGTM, but I have a general question about the props.

components/src/web-components/visualization/gs-sequences-by-location.tsx
fengelniederhammer: I'll add something to the map view doc section.

components/src/preact/map/sequences-by-location.tsx
fengelniederhammer: As discussed, we'll leave it as it is.
* the props of the web component must match the Preact props so that the example code in the info box works
* We try to have primitive props when possible, because it's easier to use in HTML

PR #610, opened 2024-12-16
fix(components): set default of `enableMapNavigation` to false

fengelniederhammer:
### Summary

https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes 
According to the HTML spec a boolean attribute can only bet set to `true` by setting it. Absence it considered as false which only works if also the default is false.
### PR Checklist


JonasKellerer (approved):
LGTM

Issue #267, opened 2024-05-23
Location filter: Improve design and user experience


chaoran-chen:
A sketch of how the location filter could potentially look like:

<img src="https://github.com/GenSpectrum/dashboards/assets/18666552/2f8a414e-3967-4a74-8719-e51dc99ef532" width="250" />

PR #622, opened 2024-12-19
feat(components): improve design of location filter

JonasKellerer:
Resolves: #267

### Summary

This is a draft of how an improved design of the location filter could look like with "react-select". It provides a simple api to create searchable select fields, which is less code to maintain. 

Open points are:
  - Is it required that users can delete single characters on their input? Currently, each time a new message must be typed. This is a shortcoming of "react-select"
  -  Do we need anything other, then the current display of the hierarchy? This is an easy solution and we would not need to provide anything other than the hierarchy to the component. Otherwise we would need the labels for each of the fields.

I will now look into [downshift](https://www.downshift-js.com/use-combobox/) and see if this will be better. I also looked into headless-ui, but this is even more code to maintain, so I would only use it, if we have to. 

### Screenshot

[Bildschirmaufzeichnung vom 19.12.2024, 09:11:13.webm](https://github.com/user-attachments/assets/01e66646-3722-4c00-84c3-328df5a4349e)

### PR Checklist


JonasKellerer:
Now it looks like this:
grafik

When the user clears the selection using the "x" button or when the user deletes all characters and then blurs, a new event is fired with an empty filter. When the user clears single charcters, but does not select a new location, and then blurs, the selection is put back to the original value.

anna-parker:
@JonasKellerer  - I love that you were able to put the x before the dropdown button - is this because you imported the style sheet? Do you know if its possible to modify the location of the label with a style sheet as well? e.g. in https://github.com/GenSpectrum/dashboard-components/pull/633 

Also if you have time it would also be great if you could explain to me how you are using react in the webcomponents as Fabian told me this wasn't possible

JonasKellerer:
Closing in favor of #638, which is the updated version using downshift.

PR #625, opened 2024-12-20
feat(components): gs-sequences-by-location: add button to download the data

### LINKED ISSUE #624: gs-prevalence-by-location: add download button

fengelniederhammer:
resolves #624

### Summary

Example download data can be seen in the added snapshots.

### PR Checklist


components/tests/snapshots.spec.ts-snapshots/Sequences-by-location-Download-of-visualizatio-da706-es-by-location--germany-should-match-snapshot-1-chromium-linux.txt
```diff
@@ -0,0 +1,20 @@
+count,division,proportion,isShownOnMap
```
JonasKellerer, line 1: More like a general question for this ticket: do we want to provide the "isShownOnMap" column in the csv data? Usually we let users download all the data that is shown on the table. However, is this data usefull in this case?

JonasKellerer (approved):
Implementation LGTM.

components/tests/snapshots.spec.ts-snapshots/Sequences-by-location-Download-of-visualizatio-da706-es-by-location--germany-should-match-snapshot-1-chromium-linux.txt
fengelniederhammer: I don't think that it's really useful, but for consistency I would include it in the downloaded file (and it simplifies the code).

PR #626, opened 2024-12-20
refactor(components): use flex-grow instead of computing the tab height in JS

fengelniederhammer:
### PR Checklist


components/src/preact/components/tabs.tsx
```diff
@@ -59,14 +40,13 @@ const Tabs: FunctionComponent<ComponentTabsProps> = ({ tabs, toolbar }) => {
...
-            <div ref={tabRef} className='flex flex-row justify-between flex-wrap'>
+        <div className='h-full w-full flex flex-col'>
+            <div className='flex flex-row justify-between flex-wrap '>
                 {tabElements}
                 {toolbar && <div className='py-2 flex flex-wrap gap-y-1'>{toolbarElement}</div>}
             </div>
             <div
-                className={`p-2 border-2 border-gray-100 rounded-b-md rounded-tr-md ${activeTab === tabs[0]?.title ? '' : 'rounded-tl-md'}`}
-                style={{ height: `calc(100% - ${heightOfTabs})` }}
+                className={`p-2 flex-grow border-2 border-gray-100 rounded-b-md rounded-tr-md ${activeTab === tabs[0]?.title ? '' : 'rounded-tl-md'}`}
```
JonasKellerer, line 49:
```suggestion
                className={`p-2 f grow border-2 border-gray-100 rounded-b-md rounded-tr-md ${activeTab === tabs[0]?.title ? '' : 'rounded-tl-md'} overflow-scroll`}
```
So "visualization-mutation-over-time--shows-message-when-too-many-mutations" is not larger that its container. Otherwise it looks like this:
grafik

components/src/preact/components/tabs.tsx
```diff
@@ -59,14 +40,13 @@ const Tabs: FunctionComponent<ComponentTabsProps> = ({ tabs, toolbar }) => {
...
             >
                 {tabs.map((tab) => (
                     <div className='h-full overflow-auto' key={tab.title} hidden={activeTab !== tab.title}>
```
JonasKellerer, line 52:
```suggestion
                    <div className='h-full' key={tab.title} hidden={activeTab !== tab.title}>
```
I think we don't need this overflow here.

JonasKellerer (approved):
Nice. Good to have this calculation removed. One minor change needs to be done, so larger elements do not escape the tab boundary, but use overflow.

Issue #609, opened 2024-12-16
Filter autocomplete options of location filter, text input and location filter


fengelniederhammer:
Extend `gs-lineage-filter `, `gs-location-filter` and `gs-text-input ` such that they accept a `LapisFilter`. The filter should be applied when fetching the list of autocomplete options (the options should only show values that are available within the already selected filter).

PR #633, opened 2025-01-02
feat(components): Filter suggestion dropdown to only include values that exist with the current filters and add show the number of options

anna-parker:
resolves https://github.com/GenSpectrum/dashboard-components/issues/609

### Summary

This shows the number of values for each option in the dropdown and filters the list of option to options that are available with current filters. 

Sadly it is not possible to style the options component so that the number of the options is next to the value and not below, we will need to create a custom component for this. Loculus does this by using `'@headlessui/react'`'s `Combobox` class -- however sadly this requires react and thus cannot be used in our webcomponents (?) I have not been able to find a good solution for this without react. 

### Screenshot

image

### PR Checklist

### DIFF
--- a/components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
+++ b/components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
@@ -1,9 +1,57 @@
 import { FetchAggregatedOperator } from '../../operator/FetchAggregatedOperator';
 
+type GroupType = Record<string, string> & { count: number };
+
 export async function fetchLineageAutocompleteList(lapis: string, field: string, signal?: AbortSignal) {
     const fetchAggregatedOperator = new FetchAggregatedOperator<Record<string, string>>({}, [field]);
 
     const data = (await fetchAggregatedOperator.evaluate(lapis, signal)).content;
 
-    return data.flatMap((item) => [item[field], `${item[field]}*`]).sort();
+    const output: (Record<string, string> & {
+        count: number;
+    })[] = [];
+
+    const findOrCreateGroup = (pattern: string) => {
+        let group = output.find((item) => item[field] === pattern) as GroupType | undefined;
+        if (!group) {
+            group = { count: 0, [field]: pattern } as GroupType;
+            output.push(group);
+        }
+        return group;
+    };
+
+    data.forEach((item) => {
+        if (!item[field]) {
+            return;
+        }
+        const parts = item[field].split('.');
+        for (let step = 0; step < parts.length; step++) {
+            const section = `${parts.slice(0, step + 1).join('.')}*`;
+            const group = findOrCreateGroup(section);
+            group.count += item.count;
+        }
+        const group = findOrCreateGroup(item[field]);
+        group.count += item.count;
+    });
+
+    return sortDataByField(output, field);
 }
+
+const sortDataByField = (data: (Record<string, string> & { count: number })[], field: string) => {
+    return data.sort((a, b) => {
+        const aValue = a[field];
+        const bValue = b[field];
+
+        if (aValue === undefined && bValue !== undefined) {
+            return 1;
+        }
+        if (bValue === undefined && aValue !== undefined) {
+            return -1;
+        }
+        if (aValue === undefined && bValue === undefined) {
+            return 0;
+        }
+
+        return aValue.localeCompare(bValue);
+    });
+};
--- a/components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
+++ b/components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
@@ -1,9 +1,15 @@
 import { FetchAggregatedOperator } from '../../operator/FetchAggregatedOperator';
+import { type LapisFilter } from '../../types';
 
 type GroupType = Record<string, string> & { count: number };
 
-export async function fetchLineageAutocompleteList(lapis: string, field: string, signal?: AbortSignal) {
-    const fetchAggregatedOperator = new FetchAggregatedOperator<Record<string, string>>({}, [field]);
+export async function fetchLineageAutocompleteList(
+    lapisFilter: LapisFilter,
+    lapis: string,
+    field: string,
+    signal?: AbortSignal,
+) {
+    const fetchAggregatedOperator = new FetchAggregatedOperator<Record<string, string>>(lapisFilter, [field]);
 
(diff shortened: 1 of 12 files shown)


components/src/preact/textInput/text-input.tsx
```diff
@@ -111,7 +121,11 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({ lapisField, pl
             </div>
             <datalist id={lapisField}>
                 {data.map((item) => (
-                    <option value={item} key={item} />
+                    <option
+                        value={item[lapisField]}
+                        key={item[lapisField]}
+                        style='white-space: nowrap;'
```
fengelniederhammer, line 127:
```suggestion
                        className='whitespace-nowrap'
```

components/src/preact/lineageFilter/lineage-filter.tsx
```diff
@@ -118,7 +121,10 @@ const LineageFilterInner: FunctionComponent<LineageFilterInnerProps> = ({
             </div>
             <datalist id={lapisField}>
                 {data.map((item) => (
-                    <option value={item} key={item} />
+                    <option
+                        value={item[lapisField]}
+                        key={item[lapisField]}
+                    >{`${item[lapisField]} (${item['count']})`}</option>
```
fengelniederhammer, line 127: Do we also want to include
```
                        className='whitespace-nowrap'
```
here?

components/src/preact/textInput/text-input.tsx
```diff
@@ -111,7 +121,11 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({ lapisField, pl
...
+                    >{`${item[lapisField]} (${item['count']})`}</option>
```
fengelniederhammer, line 128: You don't need a template string here. You can use JSX string rendering directly.
```suggestion
                    >{item[lapisField]} ({item['count']})</option>
```

components/src/preact/lineageFilter/lineage-filter.tsx
fengelniederhammer, line 127: You don't need a template string here. You can use JSX string rendering directly.
```suggestion
                    >{item[lapisField]} ({item['count']})</option>
```

components/src/preact/textInput/text-input.tsx
```diff
@@ -78,7 +88,7 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({ lapisField, pl
         if (value === undefined) {
             return true;
         }
-        return data.includes(value);
+        return data.map((item) => item[lapisField]).includes(value);
```
fengelniederhammer, line 91: It would probably be better to put `data.map((item) => item[lapisField])` in a `useMemo` so that we don't have to map the whole data on every keystroke again.

components/src/preact/textInput/fetchAutocompleteList.ts
```diff
@@ -1,9 +1,36 @@
...
+
+        if (aValue === undefined && bValue !== undefined) {
+            return 1;
+        }
+        if (bValue === undefined && aValue !== undefined) {
+            return -1;
+        }
+        if (aValue === undefined && bValue === undefined) {
+            return 0;
+        }
```
fengelniederhammer, line 32: This is shorter, less complicated, but most probably as good as the longer version ;) 
```suggestion
        if (aValue === undefined) {
            return 1;
        }
        if (bValue === undefined) {
            return -1;
        }
```

components/src/preact/textInput/fetchAutocompleteList.ts
```diff
@@ -1,9 +1,36 @@
...
 
-export async function fetchAutocompleteList(lapis: string, field: string, signal?: AbortSignal) {
-    const fetchAggregatedOperator = new FetchAggregatedOperator<Record<string, string>>({}, [field]);
+export async function fetchAutocompleteList(
+    lapisFilter: LapisFilter,
+    lapis: string,
+    field: string,
+    signal?: AbortSignal,
+) {
+    const fetchAggregatedOperator = new FetchAggregatedOperator<Record<string, string>>(lapisFilter, [field]);
```
fengelniederhammer, line 10: Actually the type `Record<string, string>` is not really correct. When you choose a field of type `int` for `field`, then the sort function won't work.

How do we deal with this? Should we generalize the sort function?

components/src/web-components/input/gs-text-input.stories.ts
```diff
@@ -14,6 +14,7 @@ import { withinShadowRoot } from '../withinShadowRoot.story';
 const codeExample = String.raw`
 <gs-text-input 
     lapisField="host"
+    lapisFilter={{}}
```
fengelniederhammer, line 17: It would be good to provide an example with more values here.

components/src/web-components/input/gs-text-input.tsx
```diff
@@ -29,6 +29,20 @@ export class TextInputComponent extends PreactLitAdapter {
     @property()
     initialValue: string | undefined = undefined;
 
+    /**
+     * A LAPIS filter to fetch the number of sequences for.
```
fengelniederhammer, line 33:
```suggestion
     * A LAPIS filter to fetch the number of sequences that is shown on each autocomplete suggestion.
```

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
```diff
@@ -1,9 +1,63 @@
...
+        for (let step = 0; step < parts.length; step++) {
+            const section = `${parts.slice(0, step + 1).join('.')}*`;
+            const group = findOrCreateGroup(section);
+            group.count += item.count;
+        }
+        const group = findOrCreateGroup(item[field]);
+        group.count += item.count;
+    });
+
+    return sortDataByField(output, field);
```
fengelniederhammer, line 43: I think this tries to do more than it can achieve:
* The sorting is off:
image
`JN.1.1` appears at the top, but `JN.1.1*` is somewhere further down. I would expect them to be next to each other.
* The computed counts for values that include sublineages are not correct, because they don't consider aliases.
* SILO will already return all possible parent lineage. IMO we should not need to compute them here.

This looks more complicated than initially thought. Let's split the changes?
* Let's discuss counts for lineages and maybe do it later.
* We can get sorting (for gs-text-input and gs-lineage-filter) and counts for gs-text-input done in this PR.

fengelniederhammer (changes requested):

fengelniederhammer:
Closing in favor of #666 :metal:

PR #638, opened 2025-01-08
feat(components): improve design of location filter

JonasKellerer:
BREAKING CHANGE: gs-location-filter prop initialValue changed to value, a json with the event value
BREAKING CHANGE: gs-location-changed event now returns null for unspecified value

Resolves: #267

### Summary

This is the second try of #622.

This now uses the downshift library to present a selectable dropdown menu. It allows for easy search, keyboard useage, delete button and has a focus on accessibility.

I chose to use the same style as before to show the hirarchy of the elements. I did not use the "... country in" section, so that the user does not have to specify the labels for the single entries. I could not use the field names, since they do not make much sense for loculus instances.

The "initialValue" prop is removed in favor of the "value" prop. This is now closer to the html input component. It also now allows to put the event detail directly as an input to the component.

Also it changes, that the event now contains null values for not specified locations. This change now allows to get the fields from the filter directly from the event.

When the user clears the selection using the "x" button or when the user deletes all characters and then blurs, a new event is fired with an empty filter. When the user clears single charcters, but does not select a new location, and then blurs, the selection is put back to the original value.

### Screenshot

grafik

### PR Checklist

### DIFF
--- a/components/package.json
+++ b/components/package.json
@@ -87,6 +87,7 @@
         "chartjs-chart-error-bars": "^4.4.0",
         "chartjs-chart-venn": "^4.3.0",
         "dayjs": "^1.11.10",
+        "downshift": "^9.0.8",
         "flatpickr": "^4.6.13",
         "gridjs": "^6.2.0",
         "leaflet": "^1.9.4",
--- a/components/src/preact/locationFilter/fetchAutocompletionList.ts
+++ b/components/src/preact/locationFilter/fetchAutocompletionList.ts
@@ -1,6 +1,10 @@
 import { FetchAggregatedOperator } from '../../operator/FetchAggregatedOperator';
 
-export async function fetchAutocompletionList(fields: string[], lapis: string, signal?: AbortSignal) {
+export async function fetchAutocompletionList(
+    fields: string[],
+    lapis: string,
+    signal?: AbortSignal,
+): Promise<Record<string, string | undefined>[]> {
     const toAncestorInHierarchyOverwriteValues = Array(fields.length - 1)
         .fill(0)
         .map((_, i) => i + 1)
@@ -20,7 +24,10 @@ export async function fetchAutocompletionList(fields: string[], lapis: string, s
             return setOfAllHierarchies;
         }, new Set());
 
-    return [...locationValues].map((json) => JSON.parse(json)).sort(compareLocationEntries(fields));
+    return [...locationValues]
+        .map((json) => JSON.parse(json))
+        .sort(compareLocationEntries(fields))
+        .map((entry) => fields.reduce((acc, field) => ({ ...acc, [field]: entry[field] ?? undefined }), {}));
 }
 
 function compareLocationEntries(fields: string[]) {
--- a/components/src/preact/locationFilter/location-filter.stories.tsx
+++ b/components/src/preact/locationFilter/location-filter.stories.tsx
@@ -1,4 +1,6 @@
-import { type Meta, type StoryObj } from '@storybook/preact';
+import { type Meta, type PreactRenderer, type StoryObj } from '@storybook/preact';
+import { expect, fn, userEvent, waitFor, within } from '@storybook/test';
+import type { StepFunction } from '@storybook/types';
 
 import data from './__mockData__/aggregated.json';
 import { LocationFilter, type LocationFilterProps } from './location-filter';
@@ -35,7 +37,7 @@ const meta: Meta<LocationFilterProps> = {
     args: {
         width: '100%',
         fields: ['region', 'country', 'division', 'location'],
-        initialValue: 'Europe',
+        value: { region: 'Europe', country: undefined, division: undefined, location: undefined },
         placeholderText: 'Enter a location',
     },
     argTypes: {
@@ -44,9 +46,9 @@ const meta: Meta<LocationFilterProps> = {
                 type: 'object',
             },
         },
-        initialValue: {
+        value: {
             control: {
-                type: 'text',
+                type: 'object',
             },
         },
         width: {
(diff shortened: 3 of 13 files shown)


components/src/preact/locationFilter/location-filter.tsx
```diff
@@ -52,71 +49,179 @@ export const LocationFilterInner = ({ initialValue, fields, placeholderText }: L
     if (error) {
         throw error;
     }
+    if (data === undefined) {
+        return <NoDataDisplay />;
+    }
```
fengelniederhammer, line 54: Is this case even possible? AFAICS `data` can only be `undefined` when `isLoading` or `error`.

components/src/preact/locationFilter/fetchAutocompletionList.ts
```diff
@@ -1,6 +1,10 @@
 import { FetchAggregatedOperator } from '../../operator/FetchAggregatedOperator';
 
-export async function fetchAutocompletionList(fields: string[], lapis: string, signal?: AbortSignal) {
+export async function fetchAutocompletionList(
+    fields: string[],
+    lapis: string,
+    signal?: AbortSignal,
+): Promise<Record<string, string | null>[]> {
```
fengelniederhammer, line 7:
```suggestion
): Promise<LapisLocationFilter[]> {
```

components/src/preact/locationFilter/location-filter.tsx
```diff
@@ -52,71 +49,179 @@ export const LocationFilterInner = ({ initialValue, fields, placeholderText }: L
...
-    const matchingEntries = Object.entries(eventDetail)
-        .filter(([, value]) => value !== undefined)
-        .reduce((filteredData, [key, value]) => filteredData.filter((it) => it[key] === value), data);
-
-    return matchingEntries.length > 0;
-};
+    return {
+        lapisFilter: locationFilter,
+        label: lastNonUndefinedValue,
+        description: `${concatenatedLocation}`,
```
fengelniederhammer, line 208:
```suggestion
        description: concatenatedLocation,
```

components/src/preact/locationFilter/location-filter.tsx
```diff
@@ -52,71 +49,179 @@ export const LocationFilterInner = ({ initialValue, fields, placeholderText }: L
...
+                divRef.current?.dispatchEvent(new LocationChangedEvent(selectedItem.lapisFilter));
+            }
+        },
+        items,
+        itemToString(item) {
+            return item?.label ?? '';
+        },
+        initialSelectedItem,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
fengelniederhammer, line 119:
```suggestion
        // @ts-expect-error -- (explain why we need this...)
```

components/src/preact/locationFilter/location-filter.stories.tsx
```diff
@@ -69,14 +71,101 @@ export const Primary: StoryObj<LocationFilterProps> = {
         <LapisUrlContext.Provider value={LAPIS_URL}>
             <LocationFilter
                 fields={args.fields}
-                initialValue={args.initialValue}
+                value={args.value}
```
fengelniederhammer, line 74: We could simply use `{...args}` instead of listing every prop individually.

components/src/web-components/input/gs-location-filter.stories.ts
```diff
@@ -199,36 +196,30 @@ export const FiresEvent: StoryObj<LocationFilterProps> = {
...
-                expect.objectContaining({
-                    detail: {
-                        region: 'Asia',
-                        country: 'Bangladesh',
-                        division: 'Rajshahi',
-                        location: 'Chapainawabgonj',
-                    },
-                }),
-            );
-        });
```
fengelniederhammer, line 232: Why did you delete this test? I think it would still be good to have a test that sets all available fields.

fengelniederhammer (changes requested):
Looks good overall!

As discussed, let's revert to fire `undefined` in the event for the unset locations.

components/src/preact/locationFilter/location-filter.tsx
```diff
@@ -52,71 +49,179 @@ export const LocationFilterInner = ({ initialValue, fields, placeholderText }: L
...
+        environment: shadowRoot,
+    });
+
+    const onInputBlur = () => {
+        if (inputValue === '') {
```
fengelniederhammer, line 124:
```suggestion
    const onInputBlur = () => {
        closeMenu();
        if (inputValue === '') {
```

Where `closeMenu` is from the `useCombobox` return value. Then the autocompletion list closes when clicking somewhere outside (which seems to be the default behavior anyway when it's not used in a shadow DOM).

components/src/preact/locationFilter/location-filter.tsx
JonasKellerer: I found a way to create an environment.

components/src/web-components/input/gs-location-filter.stories.ts
JonasKellerer: I added it back in.

anna-parker:
This is looking great! Amazing that you got this to work with storybook, I'm looking forward to adding the dropdown layout to the other filters :-) 

I just had a look though and found one weirdness - I think related to blur (reminds me of the issues I ran into with the mutation filter) - after opening the menu it stays open even if I click away from the menu

https://github.com/user-attachments/assets/bd164512-624a-4c1b-ae0e-b467befb835b

:

JonasKellerer:
> This is looking great! Amazing that you got this to work with storybook, I'm looking forward to adding the dropdown layout to the other filters :-)
> 
> I just had a look though and found one weirdness - I think related to blur (reminds me of the issues I ran into with the mutation filter) - after opening the menu it stays open even if I click away from the menu
> Screen.Recording.2025-01-10.at.02.32.32.mov
> 
> :

I fixed it. The closeMenu is now on the surrounding div.

components/src/web-components/input/gs-location-filter.tsx
```diff
@@ -14,33 +15,28 @@ import { PreactLitAdapter } from '../PreactLitAdapter';
...
+ *   district: null,
  * }
  * ```
  */
 @customElement('gs-location-filter')
 export class LocationFilterComponent extends PreactLitAdapter {
     /**
      * The initial value to use for this location filter.
-     * Must be of the form `valueForField1 / valueForField2 / ... / valueForFieldN`.
+     * Must contain values for all fields.
```
fengelniederhammer, line 36: Must it really contain values for all fields? I think we could also allow not setting all values.

That case will happen naturally, if you take the value of a change event and use it here:
* Let `event.detail` be `{ region: "Europe", country:  undefined }`
* Now if you use `<gs-location-filter value={JSON.stringify(event.detail)} ... />`
* Then it will end up as `{ "region": "Europe" }`

components/src/preact/locationFilter/location-filter.tsx
```diff
@@ -53,70 +49,180 @@ export const LocationFilterInner = ({ initialValue, fields, placeholderText }: L
...
-    return fields.reduce((acc, field, i) => ({ ...acc, [field]: fieldValues[i] }), {});
-};
+function toSelectOption(locationFilter: LapisLocationFilter, fields: string[]) {
+    const concatenatedLocation = concatenateLocation(locationFilter, fields);
 
-const hasAllUndefined = (obj: Record<string, string | undefined>) =>
-    Object.values(obj).every((value) => value === undefined);
+    const lastNonUndefinedValue = Object.values(locationFilter)
+        .filter((value) => value !== null && value !== undefined)
+        .pop();
```
fengelniederhammer, line 203: I think it would be better to go by `fields.reverse()` - there we explicitly state that the order is important. It might be a bit confusing that the order of properties in `value` is also important.

components/src/preact/locationFilter/location-filter.tsx
```diff
@@ -53,70 +49,180 @@ export const LocationFilterInner = ({ initialValue, fields, placeholderText }: L
...
+                .filter((item): item is SelectItem => item !== undefined),
+        [locationData, fields],
+    );
+
+    const initialSelectedItem = useMemo(
+        () => (value !== undefined ? toSelectOption(value, fields) : null),
+        [value, fields],
+    );
+
+    const [items, setItems] = useState(allItems.filter((item) => filterByInputValue(item, initialSelectedItem?.label)));
```
fengelniederhammer, line 84: `initialSelectedItem?.label`: "TS2345: Argument of type string | null | undefined is not assignable to parameter of type string | undefined"

fengelniederhammer (changes requested):
Let's fix those small issues, then it should be good to go :)

components/src/web-components/input/gs-location-filter.tsx
JonasKellerer: It also works without those undefined entries. I changed the docs.

components/src/preact/locationFilter/location-filter.tsx
JonasKellerer: I changed it, so we use the fields. 
FYI, fields.reverse() does not create a copy of the array, it reverses in place. Secondly, If the array is reverted (and should not be) make sure to restart storybook. It keeps the reversed entry in memory if you have the wrong implementation before.

components/package.json
JonasKellerer, line 1: Reminder: Change commit message.

components/src/preact/locationFilter/LocationFilterEvent.ts
fengelniederhammer, line 1: Let's rename the file to `LocationChangedEvent.ts`

fengelniederhammer (approved):

PR #639, opened 2025-01-08
feat(components): gs-aggregate: add bar chart view

### LINKED ISSUE #546: Aggregated: add bar chart
At the moment, we only show a table for aggregated. It would be nice to add a bar chart as well. Some thoughts:

- Horizontal bars are probably better than vertical to have more space for labels
- What if there are too many values? Should we truncate? or should it be scrollable?
- What if there is more than one field?
  - For two fields, I can imagine that having a stacked bar chart could be nice: the first value in fields will be the whole bars, and the second value used for the sub-sections of the bars
  - For more than two fields, the bar chart should simply not work?

fengelniederhammer:
resolves #546

### Summary

* We decided to filter out null values in the bar chart for now. Maybe we'll include them later.
* There is an attribute `maxNumberOfBars` that controls how many bars are shown. Maybe we can implement pagination or similar later.

(I added corresponding tasks to the parent epic)

### Screenshot

With one `field`:
image

With two `fields`:
image

### PR Checklist


components/src/preact/aggregatedData/aggregate-bar-chart.tsx
```diff
@@ -0,0 +1,166 @@
...
+            continue;
+        }
+        const yAxisKey = String(yValue);
+        const secondaryKey = String(secondaryValue);
+
+        if (!map.has(secondaryKey)) {
+            map.set(secondaryKey, []);
+        }
+        map.get(secondaryKey)?.push({
+            y: yAxisKey.toString(),
```
JonasKellerer, line 133: Isn't the yAxisKey not already of type String?

components/src/preact/aggregatedData/aggregate-bar-chart.tsx
```diff
@@ -0,0 +1,166 @@
...
+                    },
+                    tooltip: {
+                        mode: 'y',
+                        callbacks: {
+                            label: (context) => {
+                                const { x, proportion } = context.dataset.data[
+                                    context.dataIndex
+                                ] as unknown as DataPoint;
+                                return fields.length === 1
+                                    ? `${x} (${formatProportion(proportion)}})`
```
JonasKellerer, line 78:
```suggestion
                                    ? `${x} (${formatProportion(proportion)})`
```
I think there is one closing curly bracket too much.

components/src/preact/aggregatedData/aggregate-bar-chart.tsx
```diff
@@ -0,0 +1,166 @@
...
+        );
+    }
+
+    if (fields.length > 2) {
+        return (
+            <NoDataDisplay
+                message={`Cannot display a bar chart when there are more than two fields given that the data should be stratified by. Got: ${fields.join(', ')}`}
+            />
+        );
+    }
```
JonasKellerer, line 42: Should those not be an error? It feels, that the maintainer has done something wrong. Users can not change anything here, do they?

JonasKellerer: If so, we could use zod to validate the input.

components/src/preact/aggregatedData/aggregate.stories.tsx
JonasKellerer, line 1: I think we should also add the bar view to the default story here.

components/src/preact/aggregatedData/aggregate.tsx
```diff
@@ -28,6 +29,7 @@ const aggregatePropsSchema = z.object({
     pageSize: z.union([z.boolean(), z.number()]),
     width: z.string(),
     height: z.string(),
+    maxNumberOfBars: z.number(),
```
JonasKellerer, line 32: Do we want to already name this to something that we dont have to change later on, when we have pagination? I imagine that this could be pageSizeBars, or something. But maybe this is misleading for now.

JonasKellerer (changes requested):
Looks nice. 
Would it be possible, to also show the total of each stacked bar on the tooltip? Currently it shows each of the subentries. But we could also add this in another ticket.

components/src/preact/aggregatedData/aggregate-bar-chart.tsx
fengelniederhammer: Ah yes. Forgotten during the 1000 changes of this method :D

components/src/preact/aggregatedData/aggregate.tsx
fengelniederhammer: I can imagine that pagination would be a breaking change anyway, but I'm not sure. Also even with pagination, I wouldn't find `maxNumberOfBars` so bad.

But for now I would stick with this name. It's not clear whether we will introduce pagination -> let's have a non-confusing name for now.

components/src/preact/aggregatedData/aggregate-bar-chart.tsx
fengelniederhammer: I also thought about making those an error. Probably it does indeed make more sense.

We shouldn't use zod here, because those cases are valid for the table view.

fengelniederhammer:
> Would it be possible, to also show the total of each stacked bar on the tooltip?

I added it.

JonasKellerer (approved):

PR #644, opened 2025-01-09
fix(components): gs-mutations-over-time: show 0 instead of "no data" when there are sequences but no mutation

### LINKED ISSUE #635: Mutations over time: wrongly determined as "no data"?
**Case 1:** On [this page](https://genspectrum.org/mpox/single-variant?geoLocCountry=Australia&sampleCollectionDateRangeLower=2022&clade=IIb), every mutation is displayed as having only "no data":

image

**Case 2:** On [this page](https://genspectrum.org/covid/single-variant?date=2024-06-01--2024-12-16&nextcladePangoLineage=XEC*), it shows that there is no data for ORF1a:I1367L in week 27 (2024-07-01 to 2024-07-07) but [this page](https://genspectrum.org/covid/single-variant?date=2024-07-01--2024-07-07&aminoAcidMutations=ORF1a%3AI1367I&nextcladePangoLineage=XEC*) confirms that we do have coverage for the position in that week. In fact, there are three samples that do not have a mutation at the position (ORF1a:I1367I). So, for ORF1a:I1367L, we would have 0% instead of no data.

fengelniederhammer:
resolves #635 case 2

### Summary

Logic before:
If a certain time interval didn't have a certain mutation (corresponding to a cell in the mutations over time grid), we would leave the data empty -> It would display "no data", although we have sequences in that time interval.

Logic after:
If we have >0 sequences in a time interval, we fill up all empty grid cells with 0, because we have data, but the data doesn't have that mutation there. "no data" should only be shown when there is no data in that time interval in LAPIS (i.e. aggregated returns count 0).

-> A column of the grid will either have data in all cells or the whole column will show "no data".

### PR Checklist


components/src/query/queryMutationsOverTime.spec.ts
```diff
@@ -734,6 +741,139 @@ describe('queryMutationsOverTime', () => {
...
+
+        const { mutationOverTimeData } = await queryMutationsOverTimeData({
+            lapisFilter,
+            sequenceType: 'nucleotide',
+            lapis: DUMMY_LAPIS_URL,
+            lapisDateField: dateField,
+            granularity: 'day',
+        });
+
+        expect(mutationOverTimeData.getAsArray()).to.deep.equal([
```
anna-parker, line 821: sorry I find this test quite confusing - you have three mocked requests above which seem to always include a range of two days and then the two days individually (why is the range required? - I'm also confused because the numbers for the second mocked range don't add up to the two individual numbers)

components/src/preact/mutationsOverTime/__mockData__/aminoAcidMutationsByDay.ts
anna-parker, line 1: Are the changes in this file reflected in any of tests - I couldn't see anything and was curious

components/src/query/queryMutationsOverTime.ts
```diff
@@ -244,9 +245,14 @@ export function groupByMutation(
     });
```
anna-parker, line 245: Just as a general comment I'm surprised the dates require sorting, every time series object/package I have used in the past has always had sorted dates. Time series objects normally require a start time, time interval and then data points. I just wanted to point this out as unnecessary sorting can become quite inefficient and I know that the mutations over time component takes long to load.

components/src/query/queryMutationsOverTime.spec.ts
fengelniederhammer: > you have three mocked requests above which seem to always include a range of two days and then the two days individually (why is the range required?

That's how the component works:
* it queries the mutations over the whole time frame that was selected by the users
  * This is used to determine which mutations exist
  * It is also used to filter the mutations by proportion (because that filter is supposed to be on the proportion of the whole time frame rather than on the individual intervals)
* Then it queries the proportion for every individual interval, because that's the number that's shown in the component.



> I'm also confused because the numbers for the second mocked range don't add up to the two individual numbers

It's mock data - it doesn't need to make 100% sense (IMO).

components/src/preact/mutationsOverTime/__mockData__/aminoAcidMutationsByDay.ts
fengelniederhammer: Those are the numbers that are used in the Web Component story book so that we get stable numbers for the screenshots in the snapshot tests.

components/src/query/queryMutationsOverTime.ts
fengelniederhammer: I'm not sure about this one.
* This component is slow - but sorting an array of at max 200 entries quite certainly isn't the bottleneck.
* The dates should already be sorted by the way we generate them and since `Array.map` preserves the order. We could probably just skip sorting here. The order here is the order that will finally be displayed, one could also argue for "better safe than sorry".

JonasKellerer: I think it is good that we sort them here, since we dont show the dates on each table column (only on hover). If they are for some reason not sorted, we will have a difficult time to find the error.
And I agree that the sorting of the dates is not the bottleneck, the mapping of the mutations is.

JonasKellerer (approved):
Good find.

PR #659, opened 2025-01-14
feat(components): use downshift for textinput filter

anna-parker:
resolves #

### PR Checklist


components/src/preact/textInput/TextChangedEvent.ts
```diff
@@ -0,0 +1,11 @@
+type LapisTextFilter = Record<string, string | null | undefined>;
+
+export class TextChangedEvent extends CustomEvent<LapisTextFilter> {
```
fengelniederhammer, line 3:
```suggestion
export class TextInputChangedEvent extends CustomEvent<LapisTextFilter> {
```
since the event is called `gs-text-input-changed`.

components/src/preact/textInput/text-input.tsx
```diff
@@ -55,43 +55,149 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({ lapisField, pl
...
+        itemToString(item) {
+            return item || '';
+        },
+        initialSelectedItem: initialValue,
+        environment,
+    });
+
+    const onInputBlur = () => {
+        if (inputValue === '') {
+            divRef.current?.dispatchEvent(new TextChangedEvent({ [lapisField]: null }));
```
anna-parker, line 133: then this needs to be changed to 
```suggestion
            divRef.current?.dispatchEvent(new TextChangedEvent({ [lapisField]: undefined }));
```

components/src/preact/textInput/text-input.tsx
```diff
@@ -55,43 +55,149 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({ lapisField, pl
...
+            selectItem(null);
+        } else if (inputValue !== selectedItem) {
+            setInputValue(selectedItem || '');
+        }
     };
 
+    const clearInput = () => {
+        divRef.current?.dispatchEvent(new TextChangedEvent({ [lapisField]: null }));
```
anna-parker, line 141:
```suggestion
        divRef.current?.dispatchEvent(new TextChangedEvent({ [lapisField]: undefined }));
```

components/src/web-components/input/gs-text-input.stories.ts
```diff
@@ -121,21 +121,52 @@ export const FiresEvent: StoryObj<Required<TextInputProps>> = {
 
         await step('Empty input', async () => {
             await userEvent.type(inputField(), '{backspace>9/}');
+            await inputField().blur();
             await expect(listenerMock.mock.calls.at(-1)![0].detail).toStrictEqual({
-                host: undefined,
+                host: null,
```
anna-parker, line 126:
```suggestion
                host: undefined,
```

components/src/preact/textInput/text-input.stories.tsx
```diff
@@ -85,13 +85,30 @@ export const WithInitialValue: StoryObj<TextInputProps> = {
...
             expect(input).toHaveValue('Homo sapiens');
         });
+
+        await step('Remove initial value', async () => {
+            await fireEvent.click(canvas.getByRole('button', { name: 'clear selection' }));
+
+            await expect(changedListenerMock).toHaveBeenCalledWith(
+                expect.objectContaining({
+                    detail: {
+                        host: null,
```
anna-parker, line 107:
```suggestion
                        host: undefined,
```

components/src/preact/textInput/text-input.stories.tsx
```diff
@@ -85,13 +85,30 @@ export const WithInitialValue: StoryObj<TextInputProps> = {
...
+        await step('Setup event listener mock', async () => {
+            canvasElement.addEventListener('gs-text-input-changed', changedListenerMock);
+        });
+
         await waitFor(() => {
             const input = canvas.getByPlaceholderText('Enter a host name', { exact: false });
             expect(input).toHaveValue('Homo sapiens');
         });
+
+        await step('Remove initial value', async () => {
```
JonasKellerer, line 101: I would put this into its own test. When one looks at the name of the test, one would expect to see an input value.

JonasKellerer (changes requested):
Nice. Looks better now.

Mostly small things, and the change from null to undefined.

I think we can extract our custom dropdown into a component, and reuse it here, for the location filter, and the lineage filter. But we can do this also in a spearate PR. I imagine that we could provide a display function, a filter function, and an event function to the new component.

components/src/preact/textInput/text-input.stories.tsx
anna-parker: To remove the initial value I anyways have to first wait for the initialValue to be visible on the page - so maybe I can just rename this test as it will include both steps

components/src/preact/textInput/text-input.tsx
```diff
@@ -55,43 +55,146 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({ lapisField, pl
...
+            setItems(data.filter((item) => filterByInputValue(item, inputValue)));
+        },
+        onSelectedItemChange({ selectedItem }) {
+            if (selectedItem !== null) {
+                divRef.current?.dispatchEvent(new TextInputChangedEvent({ [lapisField]: selectedItem }));
+            }
+        },
+        items,
+        itemToString(item) {
+            return item || '';
```
JonasKellerer, line 122:
```suggestion
            return item ?? '';
```
I think this is the ideomatic way to write this, since the ?? operator only checks for null and undefined, and leaves all other null values like empty thing, 0, and booleans.

components/src/preact/textInput/text-input.tsx
```diff
@@ -55,43 +55,146 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({ lapisField, pl
...
+        initialSelectedItem: initialValue,
+        environment,
+    });
+
+    const onInputBlur = () => {
+        if (inputValue === '') {
+            divRef.current?.dispatchEvent(new TextInputChangedEvent({ [lapisField]: undefined }));
+            selectItem(null);
+        } else if (inputValue !== selectedItem) {
+            setInputValue(selectedItem || '');
```
JonasKellerer, line 133:
```suggestion
            setInputValue(selectedItem ?? '');
```

components/src/web-components/input/gs-text-input.stories.ts
```diff
@@ -97,7 +97,7 @@ export const Default: StoryObj<Required<TextInputProps>> = {
     },
 };
 
-export const FiresEvent: StoryObj<Required<TextInputProps>> = {
+export const FiresEventNoValidInput: StoryObj<Required<TextInputProps>> = {
```
JonasKellerer, line 100: I think it would be nice, to have these steps from "FiresEventNoValidInput" and "FiresEventValidInput" still in one Story. Thie web-components storybook should also be used as a documentation for our "maintainer users". Currently, we only have one Story for the other components, to indicate that we also fire events. So I would keep it like this, but I would also be open how to do this better.

JonasKellerer (approved):
LGTM.

Issue #660, opened 2025-01-17
Extract dropdown from text-input and location filter


JonasKellerer:
We now use downshift to render the dropdown. The usage is similar for text-input and location filter. Extract the common parts.

PR #661, opened 2025-01-17
refactor: extract downshift combobox from location filter and text input

JonasKellerer:
Resolves: #660

BREAKING CHANGE: Rename initialValue to value for text input

### Summary

Extracts the common method to reduce code duplication and pave the road for #654

Also renames the initial value to value to be in line with the generic input component.

### PR Checklist

### DIFF
--- a/components/src/preact/components/downshift-combobox.tsx
+++ b/components/src/preact/components/downshift-combobox.tsx
@@ -1,0 +1,157 @@
+import { useCombobox } from 'downshift/preact';
+import { type ComponentChild } from 'preact';
+import { useMemo, useRef, useState } from 'preact/hooks';
+
+export function DownshiftCombobox<Item, Data>({
+    allData,
+    convertToItem,
+    value,
+    filterItemsByInputValue,
+    createEvent,
+    itemToString,
+    placeholderText,
+    formatItemInList,
+}: {
+    allData: Data[];
+    convertToItem: (data: Data) => Item | undefined;
+    value?: Data;
+    filterItemsByInputValue: (item: Item, value: string) => boolean;
+    createEvent: (item: Item | null) => CustomEvent;
+    itemToString: (item: Item | undefined | null) => string;
+    placeholderText?: string;
+    formatItemInList: (item: Item) => ComponentChild;
+}) {
+    const allItems = useMemo(
+        () =>
+            allData
+                .map((data) => {
+                    return convertToItem(data);
+                })
+                .filter((item): item is Item => item !== undefined),
+        [allData, convertToItem],
+    );
+
+    const initialSelectedItem = useMemo(
+        () => (value !== undefined ? convertToItem(value) : null),
+        [value, convertToItem],
+    );
+
+    const [items, setItems] = useState(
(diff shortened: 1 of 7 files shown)


components/src/preact/components/downshift-combobox.tsx
```diff
@@ -0,0 +1,159 @@
...
+        [allData, convertToItem],
+    );
+
+    const initialSelectedItem = useMemo(
+        () => (value !== undefined ? convertToItem(value) : null),
+        [value, convertToItem],
+    );
+
+    const [items, setItems] = useState(
+        allItems.filter((item) => filterItemsByInputValue(item, itemToString(initialSelectedItem))),
```
fengelniederhammer, line 42: Doesn't it make more sense to hand the full `initialSelectedItem` to that function? The consuming code can still decide to turn it into a string somehow.
```suggestion
        allItems.filter((item) => filterItemsByInputValue(item, initialSelectedItem)),
```

components/src/preact/components/downshift-combobox.tsx
```diff
@@ -0,0 +1,159 @@
...
+    placeholderText,
+    formatItemInList,
+}: {
+    allData: Data[];
+    convertToItem: (data: Data) => Item | undefined;
+    value?: Data;
+    filterItemsByInputValue: (item: Item, value: string) => boolean;
+    createEvent: (item: Item) => CustomEvent;
+    itemToString: (item: Item | undefined | null) => string;
+    emptyEvent: () => CustomEvent;
```
fengelniederhammer, line 22: I'm not sure whether it would actually be better, but we could merge `emptyEvent` into `createEvent` by sending `null` to `createEvent` when nothing is selected (instead of calling `emptyEvent`).

Currently it seems a bit redundant that the consumer has to provide two functions that create an event.

fengelniederhammer (approved):
LGTM :+1:

components/src/preact/components/downshift-combobox.tsx
```diff
@@ -0,0 +1,159 @@
...
+    placeholderText?: string;
+    formatItemInList: (item: Item) => JSXInternal.Element;
```
fengelniederhammer, line 24: I think this should accept any return value that can be rendered. There should actually be a type somewhere in Preact? It's usually something like `Node | string | null | undefined`.

fengelniederhammer: `ComponentChildren` might do the job.

components/src/preact/components/downshift-combobox.tsx
JonasKellerer: I think it should be like this. The idea is to filter the items based on the input from the input field. So the list changes, when the user types. The itemToString option is here necessary, since the "value" must me converted to a string first.

components/src/preact/components/downshift-combobox.tsx
JonasKellerer: Good idea. I'll change it.

components/src/preact/components/downshift-combobox.tsx
JonasKellerer: `ComponentChildren` works well.

PR #666, opened 2025-01-20
609 filter autocomplete options of location filter text input and location filter

### LINKED ISSUE #609: Filter autocomplete options of location filter, text input and location filter

JonasKellerer:
resolves #609

### Summary

Adds the option to provide a lapis filter to the text-input, location-filter and lineage-filter. The filter is applied to limit the select options.

### PR Checklist

### DIFF
--- a/components/src/preact/lineageFilter/lineage-filter.stories.tsx
+++ b/components/src/preact/lineageFilter/lineage-filter.stories.tsx
@@ -24,6 +24,7 @@ const meta: Meta = {
                         url: AGGREGATED_ENDPOINT,
                         body: {
                             fields: ['pangoLineage'],
+                            country: 'Germany',
                         },
                     },
                     response: {
@@ -34,8 +35,39 @@ const meta: Meta = {
             ],
         },
     },
+    argTypes: {
+        lapisField: {
+            control: {
+                type: 'text',
+            },
+        },
+        placeholderText: {
+            control: {
+                type: 'text',
+            },
+        },
+        value: {
+            control: {
+                type: 'text',
+            },
+        },
+        width: {
+            control: {
+                type: 'text',
+            },
+        },
+        lapisFilter: {
+            control: {
+                type: 'object',
+            },
+        },
+    },
+
     args: {
         lapisField: 'pangoLineage',
+        lapisFilter: {
+            country: 'Germany',
+        },
         placeholderText: 'Enter a lineage',
         value: 'A.1',
         width: '100%',
@@ -47,12 +79,7 @@ export default meta;
 export const Default: StoryObj<LineageFilterProps> = {
     render: (args) => (
         <LapisUrlContext.Provider value={LAPIS_URL}>
-            <LineageFilter
-                lapisField={args.lapisField}
-                placeholderText={args.placeholderText}
-                value={args.value}
-                width={args.width}
-            />
+            <LineageFilter {...args} />
         </LapisUrlContext.Provider>
     ),
     play: async ({ canvasElement, step }) => {
--- a/components/src/web-components/input/gs-lineage-filter.tsx
+++ b/components/src/web-components/input/gs-lineage-filter.tsx
@@ -14,7 +14,7 @@ import { PreactLitAdapter } from '../PreactLitAdapter';
  * Currently, it is designed to work well with Pango Lineages,
  * but it may also be used for other lineage types, if suitable.
  *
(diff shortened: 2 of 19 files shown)


components/src/web-components/input/gs-lineage-filter.tsx
```diff
@@ -14,7 +14,7 @@ import { PreactLitAdapter } from '../PreactLitAdapter';
  * Currently, it is designed to work well with Pango Lineages,
  * but it may also be used for other lineage types, if suitable.
  *
- * It fetches all available values of the `lapisField` from the LAPIS instance
+ * It fetches all available values of the `lapisField` from the LAPIS instance, within the given `lapisFilter`
```
fengelniederhammer, line 17:
```suggestion
 * It fetches all available values of the `lapisField` from the LAPIS instance within the given `lapisFilter`
```

anna-parker (approved):
lgtm!

components/src/web-components/input/gs-lineage-filter.tsx
```diff
@@ -45,6 +45,20 @@ export class LineageFilterComponent extends PreactLitAdapter {
     @property()
     lapisField = '';
 
+    /**
+     * The filter to apply to the autocomplete options.
```
fengelniederhammer, line 49:
```suggestion
     * The filter that is used to fetch the available autocomplete options.
```

components/src/web-components/input/gs-lineage-filter.tsx
```diff
@@ -45,6 +45,20 @@ export class LineageFilterComponent extends PreactLitAdapter {
...
+     */
+    @property({ type: Object })
+    lapisFilter:
+        | (Record<string, string | string[] | number | null | boolean | undefined> & {
+              nucleotideMutations?: string[];
+              aminoAcidMutations?: string[];
+              nucleotideInsertions?: string[];
+              aminoAcidInsertions?: string[];
+          })
+        | undefined = undefined;
```
fengelniederhammer, line 60: Nitpick: It would simplify the types a little if we didn't allow `undefined` and had `{}` as the default.
```suggestion
    lapisFilter:
        | (Record<string, string | string[] | number | null | boolean | undefined> & {
              nucleotideMutations?: string[];
              aminoAcidMutations?: string[];
              nucleotideInsertions?: string[];
              aminoAcidInsertions?: string[];
          }) = {};
```

components/src/web-components/input/gs-lineage-filter.tsx
```diff
@@ -45,6 +45,20 @@ export class LineageFilterComponent extends PreactLitAdapter {
     @property()
     lapisField = '';
 
+    /**
+     * The filter to apply to the autocomplete options.
+     * It must be a valid LAPIS filter object.
+     */
```
fengelniederhammer, line 51: Let's mention that it will fetch all available options (without a filter) if this is not set? Or is that obvious enough?

components/src/preact/lineageFilter/lineage-filter.stories.tsx
```diff
@@ -34,8 +35,40 @@ const meta: Meta = {
...
+        width: {
+            control: {
+                type: 'text',
+            },
+        },
+        lapisFilter: {
+            control: {
+                type: 'object',
+            },
+        },
```
fengelniederhammer, line 64: Isn't most of this already inferred correctly by Storybook?

fengelniederhammer (approved):
Looks good, maybe the docs can be improved a little (for all components)

anna-parker:
wupps I realized while using this in dashboards that we need to pass in the lapisFilter as a dependency or else lapisFilter changes will not cause a rerender and defeat the point of this change, I used this in: https://github.com/GenSpectrum/dashboards/pull/486

components/src/web-components/input/gs-lineage-filter.tsx
JonasKellerer: I changed it, so the user knows that the default already filters for everything.

components/src/preact/lineageFilter/lineage-filter.stories.tsx
JonasKellerer: Most of it, but I think it is good to be explicit. We do this for the other stories as well.

PR #676, opened 2025-01-21
feat(components): gs-text-input: show count on autocomplete options

### LINKED ISSUE #664: Show counts in autocomplete options - gs-text-input

JonasKellerer:
Resolves #664

### Summary

Shows counts on the dropdown of the gs-text-input. 

### Screenshot

Screenshot 2025-01-20 134313

### PR Checklist

### DIFF
--- a/components/src/preact/textInput/fetchStringAutocompleteList.ts
+++ b/components/src/preact/textInput/fetchStringAutocompleteList.ts
@@ -12,9 +12,23 @@ export async function fetchStringAutocompleteList({
     lapisFilter?: LapisFilter;
     signal?: AbortSignal;
 }) {
-    const fetchAggregatedOperator = new FetchAggregatedOperator<Record<string, string>>(lapisFilter ?? {}, [field]);
+    const fetchAggregatedOperator = new FetchAggregatedOperator<Record<string, string | null>>(lapisFilter ?? {}, [
+        field,
+    ]);
 
     const data = (await fetchAggregatedOperator.evaluate(lapis, signal)).content;
 
-    return data.map((item) => item[field]).sort();
+    return data
+        .map((item) => ({ count: item.count, value: item[field] }))
+        .filter((item): item is { count: number; value: string } => item.value !== null)
+        .sort((a, b) => {
+            if (a.value === null) {
+                return 1;
+            }
+            if (b.value === null) {
+                return -1;
+            }
+
+            return a.value.localeCompare(b.value);
+        });
 }
--- a/components/src/preact/textInput/text-input.tsx
+++ b/components/src/preact/textInput/text-input.tsx
@@ -68,32 +68,46 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({
     return <TextSelector lapisField={lapisField} value={value} placeholderText={placeholderText} data={data} />;
 };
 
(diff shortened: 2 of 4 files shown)


components/src/preact/textInput/text-input.tsx
```diff
@@ -68,32 +68,46 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({
...
 const TextSelector = ({
     lapisField,
     value,
     placeholderText,
     data,
 }: TextSelectorProps & {
-    data: string[];
+    data: SelectItem[];
 }) => {
+    const initialSelectedItem = data.find((canditade) => canditade.value == value);
```
fengelniederhammer, line 84:
```suggestion
    const initialSelectedItem = data.find((candidate) => canditade.value == value);
```

components/src/preact/textInput/text-input.tsx
```diff
@@ -68,32 +68,46 @@ const TextInputInner: FunctionComponent<TextInputInnerProps> = ({
...
+            itemToString={(item: SelectItem | undefined | null) => item?.value ?? ''}
             placeholderText={placeholderText}
-            formatItemInList={(item: string) => {
-                return <span>{item}</span>;
+            formatItemInList={(item: SelectItem) => {
+                return (
+                    <div className='flex gap-2'>
+                        <span>{item.value}</span>
+                        <span className='text-gray-500'>({item.count})</span>
+                    </div>
```
fengelniederhammer, line 101: I don't know whether this is actually simpler to compute for a browser (because we don't need `flex` here)
```suggestion
                    <p>
                        <span>{item.value}</span>
                        <span className='ml-2 text-gray-500'>({item.count})</span>
                    </p>
```

fengelniederhammer (approved):

fengelniederhammer (changes requested):
When I change the `lapisField` to "whoClade" then I see `TypeError: a.value is null` in the browser and the component displays an error.

components/src/preact/textInput/text-input.tsx
JonasKellerer: We can use this here. I don't really like the margin calls to position elements in a row, because this causes issues if they are put on different rows. But I think we can change it here, since we only have two elements.

JonasKellerer:
The null values where not handeled. I now filter them out and leave the treatment to #679

components/src/preact/textInput/fetchStringAutocompleteList.ts
```diff
@@ -12,9 +12,26 @@ export async function fetchStringAutocompleteList({
...
     const data = (await fetchAggregatedOperator.evaluate(lapis, signal)).content;
 
-    return data.map((item) => item[field]).sort();
+    return data
+        .map((item) => ({ count: item.count, value: item[field] }))
+        .filter((item): item is { count: number; value: string } => item.value !== null)
+        .sort((a, b) => {
+            if (a.value === null && b.value === null) {
+                return 0;
+            }
```
fengelniederhammer, line 27: Nitpick: There should always only be one null value (due to the way the query is constructed). Even if there are sevel ones, the order doesn't matter, but less code does ;)
```suggestion
```

fengelniederhammer (approved):

PR #685, opened 2025-01-23
feat(components): gs-location-filter: show counts in autocomplete options

### LINKED ISSUE #665: Show counts in autocomplete options - gs-location-filter

fengelniederhammer:
resolves #665

### Screenshot

image

### PR Checklist

### DIFF
--- a/components/src/preact/locationFilter/fetchAutocompletionList.spec.ts
+++ b/components/src/preact/locationFilter/fetchAutocompletionList.spec.ts
@@ -11,11 +11,11 @@ describe('fetchAutocompletionList', () => {
             { fields, country: 'Germany' },
             {
                 data: [
-                    { count: 0, region: 'region1', country: 'country1_1', division: 'division1_1_1' },
-                    { count: 0, region: 'region1', country: 'country1_1', division: 'division1_1_2' },
-                    { count: 0, region: 'region1', country: 'country1_1', division: null },
-                    { count: 0, region: 'region1', country: 'country1_2', division: 'division1_2_1' },
-                    { count: 0, region: 'region2', country: 'country2_1', division: null },
+                    { count: 1, region: 'region1', country: 'country1_1', division: 'division1_1_1' },
+                    { count: 2, region: 'region1', country: 'country1_1', division: 'division1_1_2' },
+                    { count: 3, region: 'region1', country: 'country1_1', division: null },
+                    { count: 4, region: 'region1', country: 'country1_2', division: 'division1_2_1' },
+                    { count: 5, region: 'region2', country: 'country2_1', division: null },
                 ],
             },
         );
@@ -27,14 +27,14 @@ describe('fetchAutocompletionList', () => {
         });
 
         expect(result).to.deep.equal([
-            { region: 'region1', country: undefined, division: undefined },
-            { region: 'region1', country: 'country1_1', division: undefined },
-            { region: 'region1', country: 'country1_1', division: 'division1_1_1' },
-            { region: 'region1', country: 'country1_1', division: 'division1_1_2' },
-            { region: 'region1', country: 'country1_2', division: undefined },
-            { region: 'region1', country: 'country1_2', division: 'division1_2_1' },
-            { region: 'region2', country: undefined, division: undefined },
-            { region: 'region2', country: 'country2_1', division: undefined },
+            { value: { region: 'region1', country: undefined, division: undefined }, count: 10 },
+            { value: { region: 'region1', country: 'country1_1', division: undefined }, count: 6 },
+            { value: { region: 'region1', country: 'country1_1', division: 'division1_1_1' }, count: 1 },
+            { value: { region: 'region1', country: 'country1_1', division: 'division1_1_2' }, count: 2 },
+            { value: { region: 'region1', country: 'country1_2', division: undefined }, count: 4 },
+            { value: { region: 'region1', country: 'country1_2', division: 'division1_2_1' }, count: 4 },
+            { value: { region: 'region2', country: undefined, division: undefined }, count: 5 },
+            { value: { region: 'region2', country: 'country2_1', division: undefined }, count: 5 },
         ]);
     });
 });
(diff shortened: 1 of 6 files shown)


components/src/web-components/input/gs-location-filter.stories.ts
```diff
@@ -219,7 +219,7 @@ export const FiresEvent: StoryObj<LocationFilterProps> = {
 
         await step('Select Asia', async () => {
             await userEvent.type(inputField(), 'Asia');
-            await userEvent.click(canvas.getByRole('option', { name: 'Asia Asia' }));
+            await userEvent.click(canvas.getByRole('option', { name: /^Asia.*Asia$/ }));
```
JonasKellerer, line 222: Why was this change necessary? The other was easier to read.

components/src/preact/locationFilter/fetchAutocompletionList.spec.ts
JonasKellerer, line 1: The test now tests many things at once. It would be nice to have a test, that the counts are added up correctly in one extra test. Also for documentation for later devs.

components/src/preact/locationFilter/fetchAutocompletionList.ts
```diff
@@ -25,26 +30,54 @@ export async function fetchAutocompletionList({
...
+        }))
         .sort(compareLocationEntries(fields))
-        .map((entry) => fields.reduce((acc, field) => ({ ...acc, [field]: entry[field] ?? undefined }), {}));
+        .map(({ value, count }) => ({
+            value: fields.reduce((acc, field) => ({ ...acc, [field]: value[field] ?? undefined }), {}),
+            count,
+        }));
 }
 
+function addValueAndAllAncestors(
```
JonasKellerer, line 53: The function name does not make sense to me. What do you mean?

components/src/preact/locationFilter/fetchAutocompletionList.ts
```diff
@@ -11,8 +16,8 @@ export async function fetchAutocompletionList({
     lapis: string;
     lapisFilter?: LapisFilter;
     signal?: AbortSignal;
-}): Promise<Record<string, string | undefined>[]> {
-    const toAncestorInHierarchyOverwriteValues = Array(fields.length - 1)
+}): Promise<LocationEntry[]> {
+    const toAncestorInHierarchyOverwriteValues = Array(fields.length)
```
JonasKellerer, line 20: This function name sounds like a function to me, but it is a value.

JonasKellerer: GPT also suggested this as a simplified version. Does it do the same?
```
const toAncestorInHierarchyOverwriteValues = fields.map((_, i) =>
  fields.slice(i + 1).reduce((acc, field) => ({ ...acc, [field]: null }), {})
);
```

components/src/preact/locationFilter/location-filter.stories.tsx
```diff
@@ -88,7 +88,7 @@ export const Primary: StoryObj<LocationFilterProps> = {
             const input = await inputField(canvas);
             await userEvent.clear(input);
             await userEvent.type(input, 'Germany');
-            await userEvent.click(canvas.getByRole('option', { name: 'Germany Europe / Germany' }));
+            await userEvent.click(canvas.getByRole('option', { name: /Germany\(\d+\) Europe \/ Germany/ }));
```
JonasKellerer, line 91: Similar to below: Why was this change necessary?

JonasKellerer (changes requested):
Nice. Thanks. Only some minor things.

components/src/web-components/input/gs-location-filter.stories.ts
fengelniederhammer: The `.*` accounts for whitespaces and the "(count)"  in between. I didn't want to include that in the test.

components/src/preact/locationFilter/fetchAutocompletionList.ts
fengelniederhammer: It adds the value and all ancestors (of the value). :)

Do you have a suggestion how to make it clearer?

components/src/preact/locationFilter/location-filter.stories.tsx
fengelniederhammer: Because the count is now also in that "option name".

I wanted to have at least one test that checks that the count is there. I decided to have that test here, in the Preact Storybook tests.

components/src/preact/locationFilter/fetchAutocompletionList.spec.ts
fengelniederhammer: I agree that it would be nice in principle - but I don't really see how to do it here without adding a lot of duplication and additional complexity of the test.

components/src/preact/locationFilter/location-filter.stories.tsx
JonasKellerer: Can't we just write the count? We have it pinned by the mock.

components/src/preact/locationFilter/fetchAutocompletionList.ts
JonasKellerer: Hm... Maybe `sumValueAndAncestors`?

components/src/preact/locationFilter/fetchAutocompletionList.ts
fengelniederhammer: I tried to improve the name.

components/src/preact/locationFilter/fetchAutocompletionList.ts
fengelniederhammer: It doesn't really sum. It adds them to the map.

JonasKellerer: Then, maybe just extend the name: "addValueAndAncestorsToMap"

JonasKellerer (approved):
LGTM

PR #698, opened 2025-01-27
fix(components): gs-mutation-filter: allow for reference genome with no genes

JonasKellerer:
### Summary

Fixes the error on the gs-mutation-filter component, when a reference genome is provided with no genes.
Its mostly a refactoring.

### PR Checklist

### DIFF
--- a/components/src/preact/mutationFilter/ExampleMutation.tsx
+++ b/components/src/preact/mutationFilter/ExampleMutation.tsx
@@ -1,0 +1,68 @@
+import { useContext } from 'preact/hooks';
+import type { FC } from 'react';
+
+import { type ReferenceGenome } from '../../lapisApi/ReferenceGenome';
+import type { SequenceType } from '../../types';
+import { ReferenceGenomeContext } from '../ReferenceGenomeContext';
+
+type ExampleMutationProps = {
+    sequenceType: SequenceType;
+    mutationType: 'substitution' | 'insertion';
+};
+
+export const ExampleMutation: FC<ExampleMutationProps> = ({ sequenceType, mutationType }) => {
+    const referenceGenome = useContext(ReferenceGenomeContext);
+
+    return <b>{getExampleMutation(referenceGenome, sequenceType, mutationType)}</b>;
+};
+
+export function getExampleMutation(
+    referenceGenome: ReferenceGenome,
+    sequenceType: SequenceType,
+    mutationType: 'substitution' | 'insertion',
+) {
+    switch (sequenceType) {
+        case 'amino acid': {
+            if (referenceGenome.genes.length === 0) {
+                return '';
+            }
+
+            const firstGene = referenceGenome.genes[0].name;
+
+            switch (mutationType) {
+                case 'substitution':
+                    return `${firstGene}:57Q`;
+                case 'insertion':
+                    return `ins_${firstGene}:31:N`;
+            }
+        }
+        // Issue of linter https://github.com/typescript-eslint/typescript-eslint/issues/3455
+        // eslint-disable-next-line no-fallthrough
+        case 'nucleotide': {
+            switch (referenceGenome.nucleotideSequences.length) {
+                case 0: {
+                    return '';
+                }
+                case 1: {
+                    switch (mutationType) {
+                        case 'substitution':
+                            return '23T';
+                        case 'insertion':
+                            return 'ins_1046:A';
+                    }
+                }
+                // Issue of linter https://github.com/typescript-eslint/typescript-eslint/issues/3455
+                // eslint-disable-next-line no-fallthrough
+                default: {
+                    const firstSegment = referenceGenome.nucleotideSequences[0].name;
+                    switch (mutationType) {
+                        case 'substitution':
+                            return `${firstSegment}:23T`;
+                        case 'insertion':
+                            return `ins_${firstSegment}:10462:A`;
+                    }
+                }
+            }
+        }
+    }
+}
--- a/components/src/preact/mutationFilter/mutation-filter-info.tsx
+++ b/components/src/preact/mutationFilter/mutation-filter-info.tsx
@@ -1,130 +1,112 @@
 import { useContext } from 'preact/hooks';
-import { type FC } from 'react';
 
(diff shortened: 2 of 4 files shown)


components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -242,11 +242,12 @@ const MutationFilterSelector: FunctionComponent<{
...
-    const segmentPrefix =
-        referenceGenome.nucleotideSequences.length > 1 ? `${referenceGenome.nucleotideSequences[0].name}:` : '';
-    const firstGene = referenceGenome.genes[0].name;
+    const nucleotideSubstitution = getExampleMutation(referenceGenome, 'nucleotide', 'substitution');
+    const nucleotideInsertion = getExampleMutation(referenceGenome, 'nucleotide', 'insertion');
+    const aminoAcidSubstitution = getExampleMutation(referenceGenome, 'amino acid', 'substitution');
+    const aminoAcidInsertion = getExampleMutation(referenceGenome, 'amino acid', 'insertion');
 
-    return `Enter a mutation (e.g. ${segmentPrefix}A123T, ins_${segmentPrefix}123:AT, ${firstGene}:M123E, ins_${firstGene}:123:ME)`;
+    return `Enter a mutation (e.g. ${[nucleotideSubstitution, nucleotideInsertion, aminoAcidSubstitution, aminoAcidInsertion].join(', ')})`;
```
fengelniederhammer, line 250: I think we should filter out the empty ones.

fengelniederhammer (changes requested):
The info has a couple of places that look weird without genes:
grafik

I think it would also be good to say that the organism doesn't have any genes defined instead of omitting the "This organism has the following genes:" part.

components/src/preact/mutationFilter/ExampleMutation.tsx
```diff
@@ -21,34 +21,49 @@ export function getExampleMutation(
...
+
+                    switch (mutationType) {
+                        case 'substitution':
+                            return `${firstGene}:57Q`;
+                        case 'insertion':
+                            return `ins_${firstGene}:31:N`;
+                    }
+                }
+            }
             return '';
```
fengelniederhammer, line 41: My IDE says: "TS7027: Unreachable code detected."

components/src/preact/mutationFilter/ExampleMutation.tsx
```diff
@@ -21,34 +21,49 @@ export function getExampleMutation(
...
+                    return '';
+                }
+                case 1: {
+                    switch (mutationType) {
+                        case 'substitution':
+                            return '23T';
+                        case 'insertion':
+                            return 'ins_1046:A';
+                    }
+                    return '';
```
fengelniederhammer, line 55: My IDE says: "TS7027: Unreachable code detected."

components/src/preact/mutationFilter/ExampleMutation.tsx
```diff
@@ -21,34 +21,49 @@ export function getExampleMutation(
...
+                    const firstGene = referenceGenome.genes[0].name;
+
+                    switch (mutationType) {
+                        case 'substitution':
+                            return `${firstGene}:57Q`;
+                        case 'insertion':
+                            return `ins_${firstGene}:31:N`;
+                    }
+                }
+            }
```
fengelniederhammer, line 40: Nitpick: IMO this would be easier to read: 
```
if (length == 0) {
  return '';
}
...
```

components/src/preact/mutationFilter/mutation-filter-info.tsx
```diff
@@ -137,25 +149,62 @@ const AminoAcidMutationsInfo = () => {
 
 const InsertionWildcards = () => {
     const referenceGenome = useContext(ReferenceGenomeContext);
-    const gene = referenceGenome.genes.length !== 0 ? referenceGenome.genes[0].name : '';
+
+    if (referenceGenome.nucleotideSequences.length === 0 && referenceGenome.genes.length === 0) {
+        return null;
+    }
```
fengelniederhammer, line 155: This case is very weird anyway:

grafik
(Maybe a loading spinner that sometimes display parts of letters?)

Maybe we should simply catch that case very early and throw a user facing error like "can't filter mutations when you don't have sequences"

components/src/preact/mutationFilter/mutation-filter-info.tsx
```diff
@@ -1,58 +1,136 @@
...
                 <InfoParagraph>
                     This organism has the following segments:{' '}
                     {referenceGenome.nucleotideSequences.map((gene) => gene.name).join(', ')}.
                 </InfoParagraph>
             )}
+            {referenceGenome.genes.length !== 0 && (
+                <InfoParagraph>
+                    This organism has the following genes: {referenceGenome.genes.map((gene) => gene.name).join(', ')}.
+                </InfoParagraph>
+            )}
```
fengelniederhammer, line 69: Let's show something like "This organism doesn't support amino acid sequences", if `referenceGenome.genes.length == 0`?

fengelniederhammer (approved):

components/src/preact/mutationFilter/mutation-filter-info.tsx
fengelniederhammer: (We could also do this in a separate PR)

components/src/preact/mutationFilter/ExampleMutation.tsx
JonasKellerer: I thought so to, but then eslint complains about `ESLint: Expected a 'break' statement before 'case'.(no-fallthrough`, which I find wierd...

JonasKellerer: I found this for eslint:
https://github.com/typescript-eslint/typescript-eslint/issues/3455

fengelniederhammer: Interesting - then let's maybe ignore the lint here with a reference to the issue?

components/src/preact/mutationFilter/mutation-filter-info.tsx
JonasKellerer: #704

PR #706, opened 2025-01-28
feat(components): make component tabs layout better suited for multiline layout

### LINKED ISSUE #675: Improve design of multiline tab bars
I think we should change our design of the tabs a little, when we want to do this. (But this does not need to be done in this PR. When the tabs are on different rows, it would be nice to have an indicator/border around the tabs, to know that they are different tabs. Similar to the highlight on hover. 

grafik

_Originally posted by @JonasKellerer in https://github.com/GenSpectrum/dashboard-components/pull/504#discussion_r1923205995_
            

Maybe we can add shorter, thinner lines on the tabs that are not currently active?

fengelniederhammer:
resolves #675

### Screenshot

Before:
grafik

After:
grafik

grafik
grafik

### PR Checklist


JonasKellerer (approved):
LGTM

PR #710, opened 2025-01-30
feat(components): gs-date-range-selector: make the component controllable from a surrounding JS app

### LINKED ISSUE #683: Investigate: why does the daterangefilter work differently than the other filters?
As discussed with @JonasKellerer when using the other filter components in dashboards the initialValue (now value) must be reset in the config after each selection in order for it to be displayed in the component. Example below uses initialLocation instead of initialValue:

```
onst newPageState = useMemo(
        () => ({
            datasetFilter: {
                location,
                dateRange,
            },
            variantFilter: toVariantFilter(variantFilterConfigState),
        }),
        [location, dateRange, variantFilterConfigState],
    );

    const currentLapisFilter = useMemo(() => {
        return view.pageStateHandler.toLapisFilter(newPageState);
    }, [newPageState])

    const currentLocationFilterConfig = {...locationFilterConfig, initialLocation: location }

    return (
        <div className='flex flex-col gap-6'>
            <div>
                <SelectorHeadline>Filter dataset</SelectorHeadline>
                <BaselineSelector
                    onLocationChange={(location) => setLocation(location)}
                    locationFilterConfig={currentLocationFilterConfig}
                    onDateRangeChange={(dateRange) => setDateRange(dateRange)}
                    dateRangeFilterConfig={dateRangeFilterConfig}
                    lapisFilter={currentLapisFilter}
                />
            </div>
````

This is not the case for the daterangefilter. In many ways the daterangefilter is easier to work with and it is odd that for example the locationfilter can send a changed signal e.g. `gs-location-filter-changed` but not show the changes in the component.

### PR DESCRIPTION
**Pull request by fengelniederhammer**
resolves #683

### Summary

Now you can supply a `value` to this component. It serves as an initial value, if the attribute doesn't change. If the attribute changes, then the component will switch to use that value.

BREAKING CHANGE: gs-date-range-selector: remove `initialValue`, `initialDateFrom`, `initialDateTo`. Use `value` instead.

### PR Checklist

### DIFF
--- a/components/src/preact/dateRangeSelector/computeInitialValues.ts
+++ b/components/src/preact/dateRangeSelector/computeInitialValues.ts
@@ -1,11 +1,9 @@
-import { type DateRangeOption } from './dateRangeOption';
+import { type DateRangeOption, type DateRangeValue } from './dateRangeOption';
 import { getDatesForSelectorValue, getSelectableOptions } from './selectableOptions';
 import { UserFacingError } from '../components/error-display';
 
 export function computeInitialValues(
-    initialValue: string | undefined,
-    initialDateFrom: string | undefined,
-    initialDateTo: string | undefined,
+    value: DateRangeValue | undefined,
     earliestDate: string,
     dateRangeOptions: DateRangeOption[],
 ): {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: components/src/preact/dateRangeSelector/computeInitialValues.ts [hunk 1]
```diff
@@ -1,32 +1,36 @@
...
+    value: DateRangeValue,
     earliestDate: string,
     dateRangeOptions: DateRangeOption[],
 ): {
     initialSelectedDateRange: string | undefined;
     initialSelectedDateFrom: Date;
     initialSelectedDateTo: Date;
 } {
-    if (isUndefinedOrEmpty(initialDateFrom) && isUndefinedOrEmpty(initialDateTo)) {
+    if (value === undefined) {
```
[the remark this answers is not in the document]
JonasKellerer: As discussed, we need more discussion in person.

components/src/preact/dateRangeSelector/date-range-selector.stories.tsx
JonasKellerer, line 1: As discussed: It would be nice to have a test, that shows, that the displayed value changes, when the props value changes.

components/src/web-components/input/gs-date-range-selector.stories.ts
JonasKellerer, line 1: The example should also be changed.

JonasKellerer (commented):
Nice. Makes it much easier to use the component.

components/src/preact/dateRangeSelector/computeInitialValues.ts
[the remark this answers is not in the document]
fengelniederhammer: This is a separate issue, so I'm resolving this.

components/src/preact/dateRangeSelector/dateRangeOption.ts
```diff
@@ -22,7 +22,18 @@ export const dateRangeOptionSchema = z.object({
...
 export type DateRangeOption = z.infer<typeof dateRangeOptionSchema>;
 
-export type DateRangeSelectOption = string | { dateFrom: string; dateTo: string };
+export const dateRangeValueSchema = z.union([
+    z.string(),
+    z.object({
+        dateFrom: z.string().date().optional(),
+        dateTo: z.string().date().optional(),
+    }),
+    z.undefined(),
```
JonasKellerer, line 31: Just nitpicking, because I thought about it:
Wouldn't it make sense, to have this schema without the undefined. And then make the "value" in [date-range-selector.stories.tsx](https://github.com/GenSpectrum/dashboard-components/pull/710/files#diff-5e35336e369d9e88b3ac1f261db3c4d19663c35f8aeb874b46e42ed75e3155da) optional? This would also convey the message, that is an optional parameter and we could get rid of `DateRangeSelectOption`.

JonasKellerer (approved):
Nice. Thank you. Good to have the test, that the display changes, when the input value changes. 

Do we want to have similar tests for the other input components?

fengelniederhammer:
> Do we want to have similar tests for the other input components?

That would probably be good (also making sure that they actually behave as we expect them to).

PR #711, opened 2025-01-31
fix(components): trim trailing slash off the `lapis` url that is provided to `gs-app`

### LINKED ISSUE #703: Allow LAPIS url with trailing slash
Currently, when supplying a LAPIS url with a trailing slash to `gs-app`, it will result in invalid LAPIS queries. Make sure that we can handle LAPIS urls with and without a trailing slash.

fengelniederhammer:
resolves #703

resolves #

### PR Checklist

### DIFF
--- a/components/src/web-components/gs-app.ts
+++ b/components/src/web-components/gs-app.ts
@@ -55,7 +55,9 @@ export class AppComponent extends LitElement {
         task: async () => {
             const lapisUrl = lapisUrlSchema.parse(this.lapis);
 
-            this.referenceGenome = await fetchReferenceGenome(lapisUrl);
+            this.referenceGenome = await fetchReferenceGenome(
+                lapisUrl.endsWith('/') ? lapisUrl.slice(0, -1) : lapisUrl,
+            );
         },
         args: () => [this.lapis],
     });
--- a/components/src/preact/LapisUrlContext.ts
+++ b/components/src/preact/LapisUrlContext.ts
@@ -1,3 +1,16 @@
 import { createContext } from 'preact';
+import { useContext } from 'preact/hooks';
 
-export const LapisUrlContext = createContext('');
+const LapisUrlContext = createContext('');
+
+export const LapisUrlContextProvider = LapisUrlContext.Provider;
+
+export const useLapisUrl = () => {
+    const lapisUrl = useContext(LapisUrlContext);
+
+    if (lapisUrl.endsWith('/')) {
+        return lapisUrl.slice(0, -1);
+    }
+
+    return lapisUrl;
+};
--- a/components/src/preact/aggregatedData/aggregate.stories.tsx
+++ b/components/src/preact/aggregatedData/aggregate.stories.tsx
@@ -4,7 +4,7 @@ import { expect, waitFor, within } from '@storybook/test';
 import aggregatedData from './__mockData__/aggregated.json';
 import { Aggregate, type AggregateProps } from './aggregate';
 import { AGGREGATED_ENDPOINT, LAPIS_URL } from '../../constants';
-import { LapisUrlContext } from '../LapisUrlContext';
+import { LapisUrlContextProvider } from '../LapisUrlContext';
 import { expectInvalidAttributesErrorMessage, playThatExpectsErrorMessage } from '../shared/stories/expectErrorMessage';
 
 const meta: Meta<AggregateProps> = {
@@ -44,9 +44,9 @@ export default meta;
 
 export const Default: StoryObj<AggregateProps> = {
     render: (args) => (
-        <LapisUrlContext.Provider value={LAPIS_URL}>
+        <LapisUrlContextProvider value={LAPIS_URL}>
             <Aggregate {...args} />
-        </LapisUrlContext.Provider>
+        </LapisUrlContextProvider>
     ),
     args: {
         fields: ['division', 'host'],
--- a/components/src/preact/aggregatedData/aggregate.tsx
+++ b/components/src/preact/aggregatedData/aggregate.tsx
@@ -1,11 +1,10 @@
 import { type FunctionComponent } from 'preact';
-import { useContext } from 'preact/hooks';
 import z from 'zod';
 
+import { useLapisUrl } from '../LapisUrlContext';
 import { AggregateTable } from './aggregate-table';
 import { type AggregateData, queryAggregateData } from '../../query/queryAggregateData';
 import { lapisFilterSchema, views } from '../../types';
-import { LapisUrlContext } from '../LapisUrlContext';
 import { CsvDownloadButton } from '../components/csv-download-button';
 import { ErrorBoundary } from '../components/error-boundary';
 import { Fullscreen } from '../components/fullscreen';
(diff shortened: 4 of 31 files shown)


components/src/web-components/gs-app.ts
```diff
@@ -55,7 +55,9 @@ export class AppComponent extends LitElement {
         task: async () => {
             const lapisUrl = lapisUrlSchema.parse(this.lapis);
 
-            this.referenceGenome = await fetchReferenceGenome(lapisUrl);
+            this.referenceGenome = await fetchReferenceGenome(
```
anna-parker, line 58: Could you explain to me why this you still need to format the lapisUrl here? Won't it be formatted every time the lapisUrl is called with `useLapisUrl();` after your changes?

anna-parker (commented):
Hi @fengelniederhammer I tried to test this locally by adding a trailing `/` here: https://github.com/GenSpectrum/dashboards/blob/main/backend/src/main/resources/application-dashboards-staging.yaml#L22 but I can't get the backend to run locally using 
`./gradlew bootRun --args='--spring.profiles.active=local-db,dashboards-prod'` - do you have time on Monday to walk me through this?

components/src/web-components/gs-app.ts
fengelniederhammer: This can't use `useLapisUrl`, because that uses a Preact context. This is a pure custom HTML component that doesn't use Preact. At this point, we didn't format the url yet, so we need to do it manually.

I did the formatting for the other components in the Preact context, because it seemed to most central place to me and it will work if we decide to publish the Preact components themselves (without their web component wrapper)

fengelniederhammer:
> Hi @fengelniederhammer I tried to test this locally by adding a trailing `/` here: https://github.com/GenSpectrum/dashboards/blob/main/backend/src/main/resources/application-dashboards-staging.yaml#L22 but I can't get the backend to run locally using `./gradlew bootRun --args='--spring.profiles.active=local-db,dashboards-prod'` - do you have time on Monday to walk me through this?

The backend config is currently broken. Just don't run the backend for now.

JonasKellerer (approved):
LGTM.
I wonder, if we can change the type of the lapis url to something builtin, which can do this for us. This would also be more speaking, when having a variable with name "lapis". But I have not quickly found anything that would do that, and this would be a larger refactoring. So I would use your solution for now.

components/src/web-components/gs-app.ts
anna-parker: Ah ok, this is a bit clearer now! I still have some follow ups if that is ok.

In the description of the gs-app it says "All other `gs-*` components must be (possibly nested) children of this component". Do we have`gs-*` components without a preact context? I was under the impression each web-component has a corresponding preact configuration but maybe I missed some.

I'm just trying to understand the data flow a bit better. Currently will each preact component be called inside the gs-app wrapper? Could we have modified the lapisUrl at the gs-app level and had this passed down to all components? (I understand your fix will work when these preact components are outside the gs-app wrapper and is preferable I am just trying to confirm my understanding of the current data flow)

anna-parker:
> The backend config is currently broken. Just don't run the backend for now.

Oh no! I couldn't find an issue for this, should I create one?

components/src/web-components/gs-app.ts
fengelniederhammer: All other web components use Preact internally. `gs-app` is the only one that doesn't. We have the adapter that makes sure that all the Preact components have the context available:
https://github.com/GenSpectrum/dashboard-components/blob/574e702cac14bc6b04633a6ba56e713fd7b51e4c/components/src/web-components/PreactLitAdapter.tsx#L50-L54

> Currently will each preact component be called inside the gs-app wrapper?

Yes, otherwise the contexts won't be available. We did that to reduce boilerplate (otherwise you would need to supply the lapis url to every component individually).

> Could we have modified the lapisUrl at the gs-app level and had this passed down to all components?

I didn't find how. Also for the preact context: You pass a value to it and it will distribute it as is...

fengelniederhammer:
> I couldn't find an issue for this, should I create one?

I never felt that it's important enough. But you're welcome to create one, I also think there is none.

components/src/web-components/gs-app.ts
anna-parker: Thanks! I think I'm finally starting to understand this! 

I am very happy to approve and merge I was just wondering would have replacing 

```
@provide({ context: lapisContext })
    @property()
    lapis: string = '';
```

with a getter and setter also have worked? Sadly I dont know know to this now without the running the backend locally with a different lapis url - how did you test your changes?

```
@provide({ context: lapisContext })
    @property()
    get lapis(): string {
        return this._lapis;
    }
    set lapis(value: string) {
        this._lapis = value.replace(/\/$/, '');
    }
    private _lapis: string = '';
  ````

fengelniederhammer:
> the type of the lapis url to something builtin

What do you mean by "something builtin"?

components/src/web-components/gs-app.ts
fengelniederhammer: It might work, but I'd still prefer to have it in the Preact context, because:

> I did the formatting for the other components in the Preact context, because it seemed the most central place to me and it will work if we decide to publish the Preact components themselves (without their web component wrapper)



> Sadly I dont know know to this now without the running the backend locally with a different lapis url - how did you test your changes?

I tested it in the React example in this repo.
Anyway, why do you need the backend to test this? You can simply edit the url in the dashboard configs and run `npm run dev`. You can also use the prod config to test this locally.

JonasKellerer:
> > the type of the lapis url to something builtin
> 
> What do you mean by "something builtin"?

Like `new URL(string_of_url)`.

components/src/web-components/gs-app.ts
anna-parker: Ah ok! So changes I make to https://github.com/GenSpectrum/dashboards/blob/main/backend/src/main/resources/application-dashboards-staging.yaml#L22 will be reflected when I run `npm run dev`? Oh thats good to know, I thought it just connected to the live backend and used the config from there!

Oh... but now I'm a bit confused - so the example already has a trailing slash in the lapis url: https://github.com/GenSpectrum/dashboard-components/blob/main/examples/React/src/App.tsx#L68, and that works when I test on main?

anna-parker: hmm... yeah when I add a trailing "/" here https://github.com/GenSpectrum/dashboards/blob/main/backend/src/main/resources/application-dashboards-staging.yaml#L22  and run the website the results are the same as prod?

fengelniederhammer:
> > > the type of the lapis url to something builtin
> > 
> > 
> > What do you mean by "something builtin"?
> 
> Like `new URL(string_of_url)`.

I see. That would indeed be a larger refactoring.

components/src/web-components/gs-app.ts
fengelniederhammer: > So changes I make to https://github.com/GenSpectrum/dashboards/blob/main/backend/src/main/resources/application-dashboards-staging.yaml#L22 will be reflected when I run `npm run dev`?

Yes (depending on which environment you use of course).




> Oh... but now I'm a bit confused - so the example already has a trailing slash in the lapis url: https://github.com/GenSpectrum/dashboard-components/blob/main/examples/React/src/App.tsx#L68, and that works when I test on main?


Yes. I also wondered what this "bug" was about. The URL looked weird when you had a trailing slash (having `//` somewhere in the middle), but everything seemed to work.

anna-parker: Sounds good! For future could you describe the invalid LAPIS queries a bit better will help me to understand the changes and review faster as I'm new to the project :-)

PR #713, opened 2025-01-31
381 auto height

JonasKellerer:
resolves #381

### Summary

In this PR, the height is changed to an optional property with no default. If the value is not provided, either the full table is displayed, or the diagram with a fixed aspect ratio is shown. 

### Screenshot

Diagram with no height provided. Defaults to fixed aspect ratio.
grafik

Table with no height provided.
grafik

### PR Checklist

- [x] All necessary documentation has been adapted.
- [x] The implemented feature is covered by an appropriate test.

PR #714, opened 2025-02-03
refactor(components): memoize some chart configs

fengelniederhammer:
### Summary

While reviewing #713 I noticed that some of the chart configs aren't memoized. I believe we can boost the performance a little by wrapping them in a `useMemo`.

### PR Checklist


JonasKellerer (approved):
Good idea. LGTM.

PR #720, opened 2025-02-04
fix(components): gs-mutation-filter: fix info wrongly showing "no nucleotide sequences" when there is only one

fengelniederhammer:
resolves #

### PR Checklist


JonasKellerer (approved):
LGTM

PR #727, opened 2025-02-07
fix(components): gs-date-range-selector `value`: make sure that strings are recognized as strings

fengelniederhammer:
### Summary

 Due to using `@property({ type: Object })`, lit would always `JSON.parse` the attribute:
 - `value="Last month"` would be invalid, because it's not a valid JSON string
 - using additional quotes like `value=""Last month""` would be valid, because that is a valid JSON string
 - `value="2021"` would be invalid, because JSON.parse would parse it into a number, which is not accepted by the zod schema

 This make sure that we treat everything as a string if it doesn't parse into an object.
 JSON strings (with the additional quotes) will still be treated as a string.

 We didn't notice this in Storybook, because due to using
 ```
 html`<gs-date-range-selector .value=${args.value}/>`
 ```
 it seems that lit directly set the properties in the class instead of passing it via text in the HTML
 (i.e. setting the attributes).
https://lit.dev/docs/templates/expressions/#property-expressions
 (The browser dev tools show `<gs-date-range-selector>` without any attributes)
### PR Checklist

### DIFF
--- a/components/src/web-components/input/gs-date-range-selector.stories.ts
+++ b/components/src/web-components/input/gs-date-range-selector.stories.ts
@@ -65,6 +65,7 @@ const meta: Meta<Required<DateRangeSelectorProps>> = {
             dateRangeOptionPresets.lastMonth,
             dateRangeOptionPresets.last3Months,
             dateRangeOptionPresets.allTimes,
+            { label: '2021', dateFrom: '2021-01-01', dateTo: '2021-12-31' },
             customDateRange,
         ],
         earliestDate: '1970-01-01',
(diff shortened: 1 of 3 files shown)


JonasKellerer:
Could you explain, why we did not see this already on the dashboards site? Is it because we use react there?

components/src/web-components/input/gs-date-range-selector.stories.ts
```diff
@@ -92,6 +93,50 @@ export const Default: StoryObj<Required<DateRangeSelectorProps>> = {
         </gs-app>`,
 };
 
+export const RenderAttributesInHtmlInsteadOfUsingPropertyExpression: StoryObj<Required<DateRangeSelectorProps>> = {
```
JonasKellerer, line 96: I don't know a good solution to this, but I wanted to raise this issue: We use this storybook also as documentation for others. The two stories, are mainly tests, with long names. Could we maybe mark them in a way, so that the enduser knows it is not for them? I could think of writing "TestRenderAttributesInHtmlInsteadOfUsingPropertyExpression"?

components/src/web-components/input/gs-date-range-selector.stories.ts
```diff
@@ -92,6 +93,50 @@ export const Default: StoryObj<Required<DateRangeSelectorProps>> = {
...
+    render: (args) =>
+        html` <gs-app lapis="${LAPIS_URL}">
+            <div class="max-w-screen-lg">
+                <gs-date-range-selector
+                    .dateRangeOptions=${args.dateRangeOptions}
```
JonasKellerer, line 101: Should we maybe use the way one would use the component in plain html?
```suggestion
                    dateRangeOptions=${JSON.stringify(args.dateRangeOptions)}
```

JonasKellerer (approved):
LGTM.

JonasKellerer:
Should we also add tests to the other components that this does not happen? Or can this not occur?

components/src/web-components/input/gs-date-range-selector.stories.ts
fengelniederhammer: Unfortunately the lit-analyzer complains about that. I don't know how to ignore that lint, that's why I used the `.` version here... Do you have a better suggestion?

JonasKellerer: Hm... Not really. The lit lint does not show in my idea... 
Maybe this is not this PR, but could be just render plain html without using the lit plugin in storybook? But it seems a lot of effort... https://storybook.js.org/docs/get-started/whats-a-story

fengelniederhammer: It also doesn't show in my IDE, `npm run lint` shows it.

fengelniederhammer:
> Should we also add tests to the other components that this does not happen? Or can this not occur?

This is due to the type of `value` being essentially `string | object`, where `string` doesn't need deserialization but `object` does. AFAIK we don't have that on any other attribute, do we?

PR #728, opened 2025-02-09
Rename to filter

### LINKED ISSUE #709: Rename `gs-date-range-selector` to `gs-date-range-filter`
All our input components are called `gs-<something>-filter`, except for the `gs-date-range-selector`. I propose to rename it to match the pattern of the other components.

pflanze:
resolves #709

### PR Checklist


pflanze:
I'm not 100% sure all of the renamed identifiers belong to the same change set, it's many of them. Please verify.

components/CHANGELOG.md
fengelniederhammer, line 1: I think we shouldn't rename anything in the changelog. At that point in time, the changelog was still correct and it will mention the renaming when this is merged.

components/src/web-components/input/gs-text-filter.stories.ts
```diff
@@ -6,26 +6,26 @@ import { withComponentDocs } from '../../../.storybook/ComponentDocsBlock';
...
     lapisFilter='{"country": "Germany"}'
     placeholderText="Enter host name"
     value="Homo sapiens"
     width="50%">
-</gs-text-input>`;
+</gs-text-filter>`;
 
-const meta: Meta<Required<TextInputProps>> = {
+const meta: Meta<Required<TextFilterProps>> = {
     title: 'Input/Text input',
```
fengelniederhammer, line 24: Could you also please rename it here?

fengelniederhammer (changes requested):
Generally looks good :+1: 

I found two smaller things and it would be good to squash this so that we only the things that are necessary to consumers in the changelog (Which is renaming `gs-date-range-selector`, `gs-text-input` and `TextFilterChangedEvent` from quickly looking over the code?)

pflanze:
Rebased and ready for re-review

fengelniederhammer (approved):
LGTM, thanks :)

pflanze:
With regards to which names changed, for gs-text-input it's these:

    gs-text-input
    gs-text-input-changed
    text-input
    text-input.stories
    TextInput
    TextInputComponent
    TextInputChangedEvent
    TextInputProps
    TextInputInnerProps

I haven't checked if all of those are public (is there a central re-export). But it will be a breaking change anyway, does it matter to specify the list?

I'm merging now because I had to hand merge twice now due to changes around or using the same identifiers.

PR #731, opened 2025-02-10
fix(components): gs-sequences-by-location: don't crash table view when there is no data

fengelniederhammer:
### Summary

The table view of `gs-sequences-by-location` crashes when LAPIS doesn't return any data for the specified filters.

### Screenshot

Before:
grafik

After:
grafik

### PR Checklist

### DIFF
--- a/components/src/preact/sequencesByLocation/sequences-by-location-table.tsx
+++ b/components/src/preact/sequencesByLocation/sequences-by-location-table.tsx
@@ -32,7 +32,7 @@ export const SequencesByLocationTable: FunctionComponent<SequencesByLocationTabl
             sort: true,
             formatter: (cell: number) => formatProportion(cell),
         },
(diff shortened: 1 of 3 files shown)


components/src/preact/sequencesByLocation/sequences-by-location-map.tsx
```diff
@@ -6,6 +6,7 @@ import { useEffect, useRef } from 'preact/hooks';
 import type { EnhancedGeoJsonFeatureProperties } from '../../query/computeMapLocationData';
 import { InfoHeadline1, InfoParagraph } from '../components/info';
 import { Modal, useModalRef } from '../components/modal';
+import { NoDataDisplay } from '../components/no-data-display';
```
JonasKellerer: I think we can now remove this import?

JonasKellerer (approved):
LGTM

PR #739, opened 2025-02-10
fix(components): omit attributes that are undefined in code examples

fengelniederhammer:
### Summary

Attributes can only be undefined if a user didn't set them.

### Screenshot

After #713,
```
                <gs-prevalence-over-time
                    numeratorFilters={JSON.stringify([numerator])}
                    denominatorFilter={JSON.stringify(denominator)}
                    lapisDateField='date'
                    granularity='day'
                ></gs-prevalence-over-time>
``` 
would yield `height='undefined'`:
grafik

### PR Checklist


components/.eslintrc.json
```diff
@@ -29,11 +29,13 @@
         "@typescript-eslint/no-non-null-assertion": "off",
         "@typescript-eslint/no-floating-promises": "error",
         "@typescript-eslint/restrict-template-expressions": "off",
+        "no-unused-vars": "off",
```
JonasKellerer, line 32: Why do we need to turn this off here? Can't we add the ignore patterns here? Or is this duplicate from the "@typescript-eslint/no-unused-vars" and we should turn this also into "error"?

fengelniederhammer: It's duplicate. I didn't want to invest even more time into linters... My guess is that we should remove `plugin:@typescript-eslint/eslint-recommended` above, because we also have `plugin:@typescript-eslint/recommended-type-checked`. But that would require some investigation.

fengelniederhammer: I changed the other one to "error" ("warn" doesn't fail the CI).

JonasKellerer (approved):
LGTM

PR #743, opened 2025-02-11
feat(components): gs-mutation-filter: show error when the reference genomes are empty

### LINKED ISSUE #704: gs-mutation-filter: Throw error, when used with reference genome with no genes and segments
Show an error, when the component is used with a reference genome, which has no genes and segments.

fengelniederhammer:
resolves #704

### Screenshot

grafik

### PR Checklist


JonasKellerer (approved):
LGTM

PR #747, opened 2025-02-11
feat(components): gs-mutations-over-time: add `displayMutations` attribute to filter the displayed mutations

### LINKED ISSUE #733: Mutations over time: filter display mutations
The maintainer should be able specify a list of mutations that should be displayed.

The mean proportion filter should still work as usual.

Example:

```
<gs-mutations-over-time
    lapisFilter='{ "pangoLineage": "JN.1*", "dateFrom": "2024-01-15", "dateTo": "2024-07-10" }'
    sequenceType="nucleotide"
    displayMutations='["20G","30T"]'
/>
```

fengelniederhammer:
resolves #733

### PR Checklist


components/src/preact/mutationsOverTime/getFilteredMutationsOverTimeData.ts
```diff
@@ -1,26 +1,42 @@
+import z from 'zod';
+
 import { type MutationOverTimeDataMap } from './MutationOverTimeData';
 import { type SubstitutionOrDeletionEntry } from '../../types';
 import { Map2dView } from '../../utils/map2d';
 import type { Deletion, Substitution } from '../../utils/mutations';
 import type { DisplayedMutationType } from '../components/mutation-type-selector';
 import type { DisplayedSegment } from '../components/segment-selector';
 
+export const displayMutationsSchema = z.array(z.string()).min(1).nullable().optional();
```
JonasKellerer, line 10: It would be nice to have this type without the optional. This way we could change the call to `getFilteredMutatioOverTimeData"  `displayMutations?: DisplayMutations`. This makes it clearer, that this only an optional value. This would then have the benefit, that this is also represented in `mutationOverTimeSchema`.

components/src/web-components/visualization/gs-mutations-over-time.tsx
```diff
@@ -87,6 +87,13 @@ export class MutationsOverTimeComponent extends PreactLitAdapterWithGridJsStyles
     @property({ type: String })
     lapisDateField: string = 'date';
 
+    /**
+     * If provided, only the given mutations will be displayed.
+     * The mutations must be provided in the exact format as they would be displayed by the component.
+     */
+    @property({ type: Array })
+    displayMutations: string[] | undefined | null = null;
```
JonasKellerer, line 95: Why do we need this type to also include null? Wouldn't undefined suffice?

components/src/preact/mutationsOverTime/getFilteredMutationsOverTimeData.ts
```diff
@@ -1,26 +1,42 @@
...
+export type DisplayMutations = z.infer<typeof displayMutationsSchema>;
+
 export function getFilteredMutationOverTimeData(
     data: MutationOverTimeDataMap,
     overallMutationData: SubstitutionOrDeletionEntry<Substitution, Deletion>[],
     displayedSegments: DisplayedSegment[],
     displayedMutationTypes: DisplayedMutationType[],
     proportionInterval: { min: number; max: number },
+    displayMutations: DisplayMutations,
```
JonasKellerer, line 19: Since we have no more parameters, I would like the parameters as object better, so there is no confusion in the call.

JonasKellerer (changes requested):
Nice. This was quick. 

The preact stories need to have also the argTypes of this new filter.

components/src/preact/mutationsOverTime/getFilteredMutationsOverTimeData.ts
fengelniederhammer: I also already thought about doing that... ;)

components/src/web-components/visualization/gs-mutations-over-time.tsx
fengelniederhammer: Actually it would probably suffice. But somehow sometimes React and Storybook manage to inject `null` in there. I've seen error where a component complains that it got `null` where only `undefined` is allowed. It happens when you try to not set the attributes.
Also allowing `null` doesn't really hurt, so I thought I could inlcude it in the type.

components/src/preact/mutationsOverTime/getFilteredMutationsOverTimeData.ts
fengelniederhammer: There is no real difference between `displayMutations?: T` and `displayMutations: T | undefined`.

How would you do it in the `mutationOverTimeSchema` then? You would again have to repeat that it's optional there. That's why I'd like to have that in a single place. IMO that's more important than having the tiny readability benefit from using `displayMutations?: T`.

components/src/web-components/visualization/gs-mutations-over-time.tsx
fengelniederhammer: As discussed, will remove `null` for now.

JonasKellerer (approved):
Looks great.

PR #757, opened 2025-02-17
fix(components): make sure to render the downshift combobox with the correct items

fengelniederhammer:
### Summary

On the dashboards, usually the location filter would update it's combobox items as follows:
* fetch the autocomplete options
* render the autocomplete options
* change the lapisFilter (e.g. by changing the selected date range)
* fetch the autocomplete options
* show the loading state, unmounting the combobox and __forgetting the old autocomplete options__
* render the new autocomplete options

https://github.com/GenSpectrum/dashboards/pull/539 had red e2e tests, because the following happened:
* fetch autocomplete options for a filter that would return no option
* change the lapisFilter before the first fetch finishes to something that would return a number of options
* the first fetch finishes and persists zero autocomplete items in a `useState`
* the second fetch finishes, but its items would not be persisted in the `useState`, because the component already has its state
* The downshift combobox would still show no options, although it got items now

Fix: make sure that the items are not a state, but get computed on every render from the input (`allItems`) and the state (`inputValue`) instead
### PR Checklist

### DIFF
--- a/components/src/preact/components/downshift-combobox.tsx
+++ b/components/src/preact/components/downshift-combobox.tsx
@@ -1,6 +1,6 @@
 import { useCombobox } from 'downshift/preact';
 import { type ComponentChild } from 'preact';
-import { useRef, useState } from 'preact/hooks';
+import { useMemo, useRef, useState } from 'preact/hooks';
 
 export function DownshiftCombobox<Item>({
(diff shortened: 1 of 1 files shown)


JonasKellerer (approved):
Good catch. LGTM

PR #767, opened 2025-02-20
refactor(components): refactor modal so that one only needs to use a single component

fengelniederhammer:
### PR Checklist


components/src/preact/components/modal.tsx
```diff
@@ -3,18 +3,24 @@ import { useRef } from 'preact/hooks';
 
 export type ModalButtonProps = {
     buttonClass?: string;
+    buttonDataTip?: string;
     modalContent: ComponentChildren;
 };
 
-export const ModalButton: FunctionComponent<ModalButtonProps> = ({ children, buttonClass = '', modalContent }) => {
+export const Modal: FunctionComponent<ModalButtonProps> = ({ children, buttonClass, buttonDataTip, modalContent }) => {
```
JonasKellerer, line 10: I'm not sure if we want to add the dataTip info to the props. This is not a property of the modal itself. When you want to have a datatip, we could pass this as a children.
```
<div className={'tooltip'} data-tip={'My data tip'}>
  The actual button text
</div>
```

components/src/preact/components/modal.tsx
```diff
@@ -3,18 +3,24 @@ import { useRef } from 'preact/hooks';
 
 export type ModalButtonProps = {
     buttonClass?: string;
```
JonasKellerer, line 5:
```suggestion
    buttonClassName?: string;
```
Nitpick.

JonasKellerer (approved):
I like the idea. I would only remove the data-tip from the props and do this on the button content.

components/src/preact/components/modal.tsx
fengelniederhammer: Good idea :+1:

PR #768, opened 2025-02-20
feat(components): add mutation annotations to `gs-mutations`

### LINKED ISSUE #766: Mutation annotations: add to `gs-mutations`
Use what was established in #753

fengelniederhammer:
resolves #766

### Screenshot

image

### PR Checklist


components/src/preact/mutations/mutations.stories.tsx
```diff
@@ -37,11 +38,30 @@ const meta: Meta<MutationsProps> = {
 
 export default meta;
 
+const mutationAnnotations = [
+    {
+        name: 'I am a mutation annotation!',
+        description: 'This describes what is special about these mutations.',
+        symbol: '🞯',
```
chaoran-chen, line 45: I see in the preview:

image

and in the code, I see:

image

Is it intended, or am I missing the character? (I'm using Mac and Chrome)

fengelniederhammer: On my Ubuntu and on my Windows it looks good in Firefox:
grafik

chaoran-chen: Interesting and unexpected! I assumed that nowadays, by default, different browsers and operating systems don't have significant differences in which characters they can display. We should keep it in mind for this feature.. Instead of something cryptic, what about using one of the following symbols: `x, #, $`?

components/src/preact/components/info.tsx
```diff
@@ -21,7 +21,7 @@ export const InfoHeadline2: FunctionComponent = ({ children }) => {
 };
 
 export const InfoParagraph: FunctionComponent = ({ children }) => {
-    return <p className='text-justify my-1'>{children}</p>;
+    return <p className='text-justify text-base font-normal my-1'>{children}</p>;
```
JonasKellerer, line 24: Why was this necessary?

components/src/preact/mutations/mutations.stories.tsx
JonasKellerer, line 1: I think it would be nice to have a test, that shows that the annotations are shown.

JonasKellerer (approved):
LGTM. Only a test would be nice to show that the annotations are used.

components/src/preact/components/info.tsx
fengelniederhammer: The annotation modals are rendered within GridJS and GridJS overwrites the default font size. This makes sure that it's "normal".

JonasKellerer: Ah. Makes sense. Thx.

PR #777, opened 2025-02-25
feat(components): add mutation annotations to 'gs-mutation-comparison'

### LINKED ISSUE #772: Mutations annotations: add to `gs-mutation-comparison`
Follow up to #753

fengelniederhammer:
resolves #772

### Screenshot

image

image

### PR Checklist

  - testing that the annotations work on the Venn tab is difficult, because we cannot properly interact with the Venn diagram. We would need to click one of the intersections there, which is hidden inside a `<canvas>` element. That is not really possible with testing library capabilities.


components/src/preact/mutationComparison/mutation-comparison-venn.tsx
```diff
@@ -105,22 +110,48 @@ export const MutationComparisonVenn: FunctionComponent<MutationComparisonVennPro
...
         return noElementSelectedMessage;
     }
 
     const values = sets.datasets[0].data[selectedDatasetIndex].values;
     const label = sets.datasets[0].data[selectedDatasetIndex].label;
-    return `${label}: ${values.join(', ')}` || '';
-}
+    return (
+        <span>
+            {label}:{' '}
```
JonasKellerer, line 145:
```suggestion
           {`${label}: `}
```
Nitpick: I think this will do the same.

JonasKellerer (approved):
LGTM

PR #779, opened 2025-02-25
fix(components): make it possible to copy an annotated mutation in Firefox

### LINKED ISSUE #778: Mutation annotations: the annotated mutations should not be a button
They should be some sort of text so that you can copy them. Firefox doesn't let me copy the text content of a button (In Chrome it works though).

fengelniederhammer:
resolves #778

### Summary

Apparently Firefox doesn't let you copy the content of a button.
### PR Checklist

### DIFF
--- a/components/src/preact/components/annotated-mutation.tsx
+++ b/components/src/preact/components/annotated-mutation.tsx
@@ -63,7 +63,7 @@ const AnnotatedMutationWithoutContext: FunctionComponent<AnnotatedMutationWithou
     );
 
     return (
-        <ButtonWithModalDialog modalContent={modalContent} modalRef={modalRef}>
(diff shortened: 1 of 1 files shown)


components/src/preact/components/annotated-mutation.tsx
```diff
@@ -63,18 +63,21 @@ const AnnotatedMutationWithoutContext: FunctionComponent<AnnotatedMutationWithou
     );
 
     return (
-        <ButtonWithModalDialog modalContent={modalContent} modalRef={modalRef}>
```
JonasKellerer, line 66: I think `buttonClassName={'select-text'}` would do the same (at least i can copy it from firefox). Does this work for you as well?

fengelniederhammer: Nice. Thanks :)

JonasKellerer (approved):

PR #780, opened 2025-02-26
feat(components): allow HTML in mutation annotation description

### LINKED ISSUE #762: Mutation annotations: make it possible to render HTML

fengelniederhammer:
resolves #762

### Screenshot

image

### PR Checklist

### DIFF
--- a/components/src/preact/components/annotated-mutation.tsx
+++ b/components/src/preact/components/annotated-mutation.tsx
@@ -56,7 +56,10 @@ const AnnotatedMutationWithoutContext: FunctionComponent<AnnotatedMutationWithou
             {mutationAnnotations.map((annotation) => (
                 <Fragment key={annotation.name}>
                     <InfoHeadline2>{annotation.name}</InfoHeadline2>
-                    <InfoParagraph>{annotation.description}</InfoParagraph>
+                    <InfoParagraph>
+                        {/* eslint-disable-next-line react/no-danger */}
+                        <div dangerouslySetInnerHTML={{ __html: annotation.description }} />
+                    </InfoParagraph>
                 </Fragment>
             ))}
         </div>
--- a/components/src/preact/components/annotated-mutation.tsx
+++ b/components/src/preact/components/annotated-mutation.tsx
@@ -1,3 +1,4 @@
+import DOMPurify from 'dompurify';
 import { useRef } from 'gridjs';
 import { Fragment, type FunctionComponent, type RefObject } from 'preact';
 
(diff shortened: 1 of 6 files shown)


components/src/preact/components/annotated-mutation.tsx
```diff
@@ -56,7 +56,10 @@ const AnnotatedMutationWithoutContext: FunctionComponent<AnnotatedMutationWithou
             {mutationAnnotations.map((annotation) => (
                 <Fragment key={annotation.name}>
                     <InfoHeadline2>{annotation.name}</InfoHeadline2>
-                    <InfoParagraph>{annotation.description}</InfoParagraph>
+                    <InfoParagraph>
+                        {/* eslint-disable-next-line react/no-danger */}
+                        <div dangerouslySetInnerHTML={{ __html: annotation.description }} />
```
JonasKellerer, line 61: This now leaves us open to XSS attacks (Assumingt that our consumers do not take care of that).
GPT suggested this:
```
import DOMPurify from 'dompurify';

const sanitizedHTML = DOMPurify.sanitize(annotation.description);

<div dangerouslySetInnerHTML={{ __html: sanitizedHTML }} />
```
With the question: "What is considered safe html?"
Safe HTML refers to HTML content that does not contain any elements, attributes, or scripts that could be exploited to perform malicious actions, such as Cross-Site Scripting (XSS) attacks. Safe HTML ensures that the content can be rendered in a browser without posing security risks to users. Here are some characteristics of safe HTML:
  - No Inline Scripts: Safe HTML does not include <script> tags or inline event handlers (e.g., onclick, onerror) that can execute JavaScript.
  - No Dangerous Attributes: Attributes that can execute scripts or cause security issues, such as javascript: URLs, are removed or sanitized.
  - No Untrusted Content: Only content from trusted sources is included, and any user-generated content is properly sanitized.
  - Properly Escaped Content: Special characters are escaped to prevent them from being interpreted as HTML or JavaScript. For example, `<` becomes `&lt;` and `>` becomes `&gt;`.
  - Allowed Tags and Attributes: Only a predefined set of HTML tags and attributes are allowed. This typically includes basic formatting tags like `<p>`, `<b>`, `<i>`, `<ul>`, `<li>`, etc., while excluding potentially harmful ones.
  - No External Resources: Links to external scripts, stylesheets, or other resources are carefully controlled or disallowed to prevent loading malicious content.

By adhering to these principles, you can ensure that HTML content is safe for rendering in a web application. Libraries like DOMPurify help automate this process by sanitizing HTML content according to these guidelines, removing or neutralizing any potentially unsafe elements.

JonasKellerer (commented):
I like the changes, but my question would be if we need to sanatize the html?

components/src/preact/components/annotated-mutation.tsx
fengelniederhammer: I also thought about that. Maybe we could simply do it to be on the safe side.

On the other hand only maintainers will put HTML there. I didn't see a way how users could inject malicious HTML in there. Might be that I missed something.

JonasKellerer: I don't think this about our security, but that of users. There an attacker could insert html inside the component and harm the user? I'm not sure about that, but that is what I though could happen.
I would prefer to be on the safe side ;)

fengelniederhammer: I'll add it.

I was wondering: How would an attacker get the malicious HTML there? It will be taken from the config that the server sends to the client.

anna-parker (approved):

PR #793, opened 2025-03-04
702 upgrade to tailwind 4

### LINKED ISSUE #702: Upgrade to Tailwind 4
I tried upgrading Tailwind to version 4 following their upgrade guide: https://tailwindcss.com/docs/upgrade-guide

The upgrade script currently fails:
```
$ npx @tailwindcss/upgrade@next --force
≈ tailwindcss v4.0.0

│ Searching for CSS files in the current directory and its subdirectories… 

│ ↳ Could not determine configuration file for: `./src/styles/tailwind.css` 
│   Update your stylesheet to use `@config` to specify the correct configuration file explicitly and then run the upgrade tool again. 

│ ↳ Could not determine configuration file for: `./node_modules/tailwindcss/base.css` 
│   Update your stylesheet to use `@config` to specify the correct configuration file explicitly and then run the upgrade tool again. 

│ ↳ Could not determine configuration file for: `./node_modules/tailwindcss/components.css` 
│   Update your stylesheet to use `@config` to specify the correct configuration file explicitly and then run the upgrade tool again. 

│ ↳ Could not determine configuration file for: `./node_modules/tailwindcss/screens.css` 
│   Update your stylesheet to use `@config` to specify the correct configuration file explicitly and then run the upgrade tool again. 

│ ↳ Could not determine configuration file for: `./node_modules/tailwindcss/tailwind.css` 
│   Update your stylesheet to use `@config` to specify the correct configuration file explicitly and then run the upgrade tool again. 

│ ↳ Could not determine configuration file for: `./node_modules/tailwindcss/utilities.css` 
│   Update your stylesheet to use `@config` to specify the correct configuration file explicitly and then run the upgrade tool again. 

│ ↳ Could not determine configuration file for: `./node_modules/tailwindcss/variants.css` 
│   Update your stylesheet to use `@config` to specify the correct configuration file explicitly and then run the upgrade tool again. 
``` 

The first issue can be solved by adding `@config "./tailwind.config.mjs";` to `./src/styles/tailwind.css`. I don't know why the others are failing.

JonasKellerer:
resolves #702

### Summary

  - Upgrades to tailwind v4, daisyui v5
  - Fixes a small alignment issue in the toolbar
  - puts the "fullscreen" button into a button, so it has the same form as the other toolbar buttons
### PR Checklist


components/.gitignore
JonasKellerer, line 1: Reminder: Squash before merge

components/src/preact/components/tabs.tsx
```diff
@@ -45,7 +45,7 @@ const Tabs: FunctionComponent<ComponentTabsProps> = ({ tabs, toolbar }) => {
                 {tabElements}
                 {toolbar && <div className='py-2 flex flex-wrap gap-y-1'>{toolbarElement}</div>}
```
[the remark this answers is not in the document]
JonasKellerer: It still works, but gap-y makes the gap vertically. I removed the x-margin on the download button, since it was inconsistent. If we want to have the button spaced out, I think we should add a gap instead. However, I would favor no gap.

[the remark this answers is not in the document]
fengelniederhammer: I'm also fine with having no gap.

components/src/preact/lineageFilter/lineage-filter.stories.tsx
```diff
@@ -123,7 +123,7 @@ export const OnBlurInput: StoryObj<LineageFilterProps> = {
     play: async ({ canvasElement, step }) => {
         const { canvas, lineageChangedListenerMock } = await prepare(canvasElement, step);
 
-        await step('after cleared selection by hand and then blur fires event with empty filter', async () => {
+        await step('after cleared selection by hand and then blur-sm fires event with empty filter', async () => {
```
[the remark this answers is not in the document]
JonasKellerer: Good catch. I found some other places, but missed this one.

components/tailwind.config.mjs
```diff
@@ -1,22 +0,0 @@
...
-    theme: {
-        extend: {
-            containers: {
-                '2xs': '18rem',
-                '3xs': '16rem',
-                '4xs': '14rem',
-            },
-        },
-    },
-    plugins: [daisyui, containerQueries, addIconSelectors(['mdi', 'mdi-light'])],
```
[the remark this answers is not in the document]
JonasKellerer: We do not need them, but I added them back in, so we have the same way as before. They only provide us with a shorthand notation like:
`iconify mdi--fullscreen-exit`. When not using this, we would need to write `icon-[mdi--fullscreen-exit]`. https://iconify.design/docs/usage/css/tailwind/tailwind4/

components/src/preact/components/color-scale-selector-dropdown.tsx
```diff
@@ -10,7 +10,7 @@ export const ColorScaleSelectorDropdown: FunctionComponent<ColorScaleSelectorDro
     setColorScale,
 }) => {
     return (
-        <div className='w-20'>
+        <div className='w-20 inline-flex'>
```
fengelniederhammer, line 13: Why do we actually need this?

components/tests/snapshots.spec.ts-snapshots/Sequences-by-location-Story-visualization-sequ-3526d-on--with-fixed-height-should-match-screenshot-1-chromium-linux.png
fengelniederhammer, line 1: A couple of countries can't be matched onto the map anymore (most notably the UK, that's why the map now only shows 50% instead of 100% of the data). Why is that? I wouldn't expect any changes there.

fengelniederhammer (changes requested):
The lineage, location and text filters don't fill the whole component anymore. Some div inside seems to be smaller now:

Before:
image

After:
image

fengelniederhammer:
I spent some time to figure out why we actually need to include the style in pages that use the components. It's weird because then the styles are duplicate: The `PreactLitAdapter` already includes them.

I didn't fully understand why we need it. My findings so far:
* The flatpickr date pickers are rendered outside the shadow dom (I don't know why though). Thus, they will be completely unstyled if the light dom doesn't have the component styles.
* Some DaisyUI CSS variables seem to be missing if you don't include the component styles. I don't know why though.

components/src/preact/components/color-scale-selector-dropdown.tsx
JonasKellerer: This is the same issue as with the help button: The surrounding div is larger as the content and we have these alignment issues.
grafik

 I am not entirely sure why, but I think this is connected to the line height. This makes it, that we fit the content.

components/tests/snapshots.spec.ts-snapshots/Sequences-by-location-Story-visualization-sequ-3526d-on--with-fixed-height-should-match-screenshot-1-chromium-linux.png
JonasKellerer: Good find, but I think this was not introduced in this PR. When I look at the currently deployed components, they also have only 53% of the data matched. I've added a ticket. #797

JonasKellerer:
I also looked into how flatpickr is able to render the date selector outside of the shadow root, because I thought it would be nice to have the dropdowns not changing the overflow behavior. There is an option for flatpickr, to render the date selector directly next to the input div. But I am not sure, if we want that. Maybe we find a solutions with the new popover feature of html, together with floating-ui (which we already use). https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover

PR #814, opened 2025-03-17
feat(components): add a genome-data-viewer

anna-parker:
resolves https://github.com/GenSpectrum/dashboards/issues/581

### Summary

1. Read in a gff3 file, create a list of CDS with name, start and end positions
2. Use this to create a CDS gene viewer 

### Screenshot

https://github.com/user-attachments/assets/805bf327-b545-4e96-926b-5932781ec283

### PR Checklist

- [x] Handle case where genes overlap
- [x] Handle case where gene has gaps


components/src/preact/genomeViewer/CDSPlot.tsx
```diff
@@ -0,0 +1,248 @@
...
+    zoomStart: number;
+    zoomEnd: number;
+}
+
+const CDSBars: FunctionComponent<CDSBarsProps> = (componentProps) => {
+    const { gffData, zoomStart, zoomEnd } = componentProps;
+    const visibleRegionLength = useMemo(() => zoomEnd - zoomStart, [zoomStart, zoomEnd]);
+    return (
+        <>
+            {gffData.map((data, list_id) => (
```
fengelniederhammer, line 153:
```suggestion
            {gffData.map((data, listId) => (
```

components/src/preact/genomeViewer/CDSPlot.tsx
```diff
@@ -0,0 +1,248 @@
...
+                                        position={tooltipPosition}
+                                        key={`${idx}-${posIdx}`}
+                                        tooltipStyle={{ left: `${leftPercent}%` }}
+                                    >
+                                        <div
+                                            className='absolute text-xs text-white rounded px-1 py-0.5 cursor-pointer hover:opacity-80 shadow-md'
+                                            style={{
+                                                left: `${leftPercent}%`,
+                                                width: `${widthPercent}%`,
+                                                backgroundColor: singleGraphColorRGBByName(cds.color!),
```
fengelniederhammer, line 181: Actually it would be good if we didn't need the `!` here. Is it possible to improve the types such that we know that the values is present here already from the typesß

components/src/preact/genomeViewer/CDSPlot.tsx
```diff
@@ -0,0 +1,248 @@
...
+    return (
+        <div className={'w-full h-6 relative'}>
+            {ticks.map((tick, idx) => {
+                const width = tick.end - tick.start;
+                const widthPercent = (width / visibleRegionLength) * 100;
+                const leftPercent = ((tick.start - zoomStart) / visibleRegionLength) * 100;
+                return (
+                    <div
+                        key={idx}
+                        class='absolute text-xs text-black px-1 py-0.5 cursor-pointer hover:opacity-80 border-l border-r border-gray-400 border-t'
```
fengelniederhammer, line 99: Why does this have `cursor-pointer`? I find it confusing because I would assume that something happens when I click it.

components/src/preact/genomeViewer/CDSPlot.tsx
```diff
@@ -0,0 +1,248 @@
...
+                                    >
+                                        <div
+                                            className='absolute text-xs text-white rounded px-1 py-0.5 cursor-pointer hover:opacity-80 shadow-md'
+                                            style={{
+                                                left: `${leftPercent}%`,
+                                                width: `${widthPercent}%`,
+                                                backgroundColor: singleGraphColorRGBByName(cds.color!),
+                                                whiteSpace: 'nowrap',
+                                                overflow: 'hidden',
+                                                textOverflow: 'ellipsis',
```
fengelniederhammer, line 184: I think it should be possible to specify those via Tailwind?

components/src/preact/genomeViewer/CDSPlot.tsx
```diff
@@ -0,0 +1,248 @@
...
+                                    return null;
+                                }
+
+                                const widthPercent = ((end - start) / visibleRegionLength) * 100;
+                                const leftPercent = ((start - zoomStart) / visibleRegionLength) * 100;
+                                const tooltipPosition = getTooltipPosition(start, end, visibleRegionLength);
+                                const tooltipContent = getToolTipContent(cds);
+                                return (
+                                    <Tooltip
+                                        content={tooltipContent}
```
fengelniederhammer, line 171: This is actually a component -> let's make it one: rename `getToolTipContent` -> `TooltipContent` and make the arguments of that function `props`, and:
```suggestion
                                
                                return (
                                    <Tooltip
                                        content={<TooltipContent ... />}
```

fengelniederhammer (changes requested):

components/src/preact/genomeViewer/CDSPlot.tsx
anna-parker: Ah I just reused the class from the CDS bars - I will remove

components/src/web-components/visualization/gs-genome-data-viewer.tsx
```diff
@@ -0,0 +1,92 @@
...
+    @property({ type: String })
+    gff3Source!: string;
+
+    /**
+     * Required
+     *
+     * The source of the gff3 file. See component level docs for more information.
+     */
+    @property({ type: Number })
+    genomeLength!: number;
```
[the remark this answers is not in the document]
anna-parker: Oh I asked chatgpt - in my understanding it asserts that these arguments are required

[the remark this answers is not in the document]
fengelniederhammer: Are you sure? From my Typescript understanding I would guess that it asserts that the value is set, even when it isn't (which is the case by default here). What happens when you remove them?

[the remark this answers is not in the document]
anna-parker: If I dont add it I get a warning that I need to specify a default, which I dont want to do - I want the user to always specify the gff3 and length: 
```
Property 'gff3Source' has no initializer and is not definitely assigned in the constructor
```

components/src/preact/genomeViewer/CDSPlot.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+function getToolTipContent(cds: CDSFeature) {
+    const cdsLength = cds.positions.reduce((acc, pos) => acc + pos.end - pos.start, 0);
+    const cdsCoordinates = cds.positions.map((pos) => `${pos.start}-${pos.end}`).join(', ');
+    return (
+        <div>
+            <p>
+                <span className='font-bold'>{cds.label}</span>
+            </p>
+            <tbody>
```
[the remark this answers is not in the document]
anna-parker: I hope this should fix the error - it is now wrapped as a table

components/src/web-components/visualization/gs-genome-data-viewer.tsx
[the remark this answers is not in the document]
fengelniederhammer: Ah ok. Then let's validate using zod.

In the end, this is an HTML tag. HTML tags are completely valid (syntactically) when you don't supply any attribute. That's why in web components, you cannot really have "required" attributes. They always need a default, thus we should put one here to satisfy the typechecker.

I would propose to add `''` and `0` as default values and then require in the zod schema that `gff3Source` is not empty and that `genomeLength > 0`.

components/src/preact/genomeViewer/CDSPlot.tsx
anna-parker: Maybe this is easier than I think. But Im not sure how to do this without creating a separate type for `CDSFeature` where color cannot be undefined or assigning a color before the CDS is fully sorted in `getCDSMap`. Which I both find suboptimal - how did you have this in mind?

fengelniederhammer: I'd prefer the separate type. Proper types are very important in the long run.

components/src/preact/genomeViewer/CDSPlot.tsx
```diff
@@ -0,0 +1,254 @@
...
+        case 'vh':
+            normalizedRatio = value / 100;
+            break;
+        default:
+            normalizedRatio = 1;
+    }
+
+    const ticks = Math.round(MAX_TICK_NUMBER * normalizedRatio);
+    return Math.max(MIN_TICK_NUMBER, Math.min(MAX_TICK_NUMBER, ticks));
+}
```
fengelniederhammer, line 44: I think the logic based on the width attribute doesn't really do the job. Usually we use our components with `width="100%"` and restrict the width with surrounding divs. 

It yields different results for the same width:
`width=400px`:
image

`width=100%`:
image

I don't know what the best approach would be.
Maybe we could get one of the top level divs (using a `useRef`) and then get the size of the HTML element.
https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect
That would give us the actual size of the component and spare us from decomposing the width based on its units.

components/src/preact/components/tooltip.tsx
```diff
@@ -39,12 +41,13 @@ function getPositionCss(position?: TooltipPosition) {
     }
 }
 
-const Tooltip: FunctionComponent<TooltipProps> = ({ children, content, position = 'bottom' }) => {
+const Tooltip: FunctionComponent<TooltipProps> = ({ children, content, position = 'bottom', tooltipStyle }) => {
     return (
-        <div className='relative w-full h-full'>
-            <div className='peer w-full h-full'>{children}</div>
+        <div className={`relative group`}>
+            <div>{children}</div>
```
fengelniederhammer, line 47: Why did you change this from peer to group?

I don't know whether it's related to this, but the layout of the mutations over time tooltips changed:
Before:
image

Now:
image

fengelniederhammer (changes requested):
Actually it looks good, thanks for the implementation!

The code needs a bit more refinement though (usability for users, robustness, refactoring). Feel free to ignore the nitpicks (I just noted down everything that came to my mind).

components/src/preact/genomeViewer/loadGff3.tsx
```diff
@@ -0,0 +1,144 @@
...
+function getCDSMap(lines: string[], genome_type: string, geneMap: CDSMap): CDSMap {
+    for (const line of lines) {
+        if (line.startsWith('#') || line.trim() === '') {
+            continue;
+        }
+
+        const fields = line.split('\t');
+        if (fields.length < 9) {
+            continue;
+        }
```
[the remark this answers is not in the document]
anna-parker: There is a reason to not throw an error. Often gff3 files start with comments - these do not have all columns so we cannot throw an error in these cases

components/src/preact/genomeViewer/loadGff3.tsx
```diff
@@ -0,0 +1,144 @@
...
+        if (type.toLowerCase() !== genome_type.toLowerCase()) {
+            continue;
+        }
+
+        const start = parseInt(startStr, 10);
+        const end = parseInt(endStr, 10);
+
+        const attrPairs = getAttributes(attributes);
+        const id =
+            attrPairs.get('ID') || attrPairs.get('Name') || attrPairs.get('gene') || attrPairs.get('gene_name') || '';
```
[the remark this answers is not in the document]
anna-parker: I think it does matter - my understanding is in a different case the attributes are invalid

components/src/preact/genomeViewer/loadGff3.tsx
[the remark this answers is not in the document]
fengelniederhammer: For my understanding: Aren't comments already filtered out by ignoring lines that start with `#`?

components/src/preact/genomeViewer/loadGff3.tsx
```diff
@@ -0,0 +1,144 @@
...
+        const sortedPositions = feature.positions.sort((a, b) => a.start - b.start);
+        mapValues[i].positions = sortedPositions;
+    }
+
+    mapValues.sort((a, b) => {
+        return a.positions[0].start - b.positions[0].start;
+    });
+
+    const GraphColorList: GraphColor[] = Object.keys(ColorsRGB) as GraphColor[];
+    let colorIndex = mapValues[0].label[0].toUpperCase().charCodeAt(0);
```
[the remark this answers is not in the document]
anna-parker: I changed this because otherwise the genome viewers always start with the same color - for multi-segmented viruses (e.g. CCHF) this can mean that all genes have the same color which looks ugly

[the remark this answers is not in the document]
fengelniederhammer: Ok, a funny but valid reason 👍

components/src/preact/genomeViewer/loadGff3.tsx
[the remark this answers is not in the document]
fengelniederhammer: Your code also already ignores blank lines (which is good I suppose).

components/src/preact/genomeViewer/CDSPlot.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+    return (
+        <div class='p-4 relative'>
+            <h2 class='text-xl font-semibold mb-2'>Genome Data Viewer</h2>
+            <div class='relative w-full'>
+                <CDSBars gffData={gffData} zoomStart={zoomStart} zoomEnd={zoomEnd} />
+                <XAxis zoomStart={zoomStart} zoomEnd={zoomEnd} fullWidth={size.width} />
+            </div>
+            <div className='h-10' />
+            <div class='relative w-full'>
```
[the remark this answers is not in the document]
anna-parker: I had to restructure the x-axis components to do this - the minmaxrangeslider is still outside its parent component tho

components/src/web-components/visualization/gs-genome-data-viewer.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { GenomeDataViewer } from '../../preact/genomeViewer/genome-data-viewer';
+import { PreactLitAdapter } from '../PreactLitAdapter';
+
+/**
+ * ## Context
+ *
+ * This component shows the Coding Sequence (CDS) of a genome using a gff3 file as input.
+ * The CDS shows which parts of the genome are translated into proteins.
+ *
+ */
```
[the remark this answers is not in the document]
anna-parker: we are using the same gff3 format that nextclade uses - so I think we should just link to that.

components/src/web-components/visualization/gs-genome-data-viewer.tsx
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * The source of the gff3 file. See component level docs for more information.
+     */
+    @property({ type: String })
+    gff3Source: string = '';
+
+    /**
+     * Required
+     *
+     * The source of the gff3 file. See component level docs for more information.
```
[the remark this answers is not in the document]
anna-parker: I added an option to get it from the gff3 file - but some nextclade gff3 datasets dont have this info so there it will need to be added as input

components/src/preact/genomeViewer/CDSPlot.tsx
anna-parker: I tried that out - it works quite nicely! Thanks for the suggestion

components/src/web-components/visualization/gs-genome-data-viewer.stories.ts
```diff
@@ -0,0 +1,75 @@
...
+        </gs-app>
+    `,
+};
+
+export const Default: StoryObj<Required<GenomeDataViewerProps>> = {
+    ...Template,
+    args: {
+        genomeLength: 11029,
+        gff3Source:
+            'https://raw.githubusercontent.com/nextstrain/nextclade_data/8f2e791d3a59013ee88e1d1d7e83b486d39c4ecb/data/nextstrain/wnv/all-lineages/genome_annotation.gff3',
```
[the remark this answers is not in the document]
anna-parker: Ok I just spent almost 2hours trying to debug these playwright tests - I cannot figure out why I can find the text if it is in an header but not text in the CDS div - I feel like I have tried everything and nothing works - if you have a solution I would be very greatful.

components/src/web-components/visualization/gs-genome-data-viewer.tsx
```diff
@@ -44,7 +44,27 @@ export class GenomeDataViewerComponent extends PreactLitAdapter {
...
+        window.removeEventListener('resize', this.updateSize);
+    }
+
+    override firstUpdated() {
+        this.updateSize(); // Get initial size after first render
+    }
+
+    private updateSize = () => {
+        this.rect = this.getBoundingClientRect(); // Measure the actual component size
+    };
```
fengelniederhammer, line 67: Could you please move this logic to the Preact component? We try to keep the interface of web components and Preact components equal.

anna-parker:
@fengelniederhammer I am having some issues with the screenshot tests - when I run the action to add them the action fails, the tests also fail... do you know how to fix this? I cant run the action locally sadly

fengelniederhammer:
> @fengelniederhammer I am having some issues with the screenshot tests - when I run the action to add them the action fails, the tests also fail... do you know how to fix this? I cant run the action locally sadly

I reran it, looks like it worked now.
It didn't find anything to commit, that's why it failed on the previous run. Maybe you ran it on an old version of the branch?

anna-parker:
> I reran it, looks like it worked now.
It didn't find anything to commit, that's why it failed on the previous run. Maybe you ran it on an old version of the branch?

Odd - I was pretty sure I ran it (twice) on the latest version, but maybe it was a github issue - thanks for fixing this now1

components/src/web-components/visualization/gs-genome-data-viewer.stories.ts
```diff
@@ -0,0 +1,112 @@
...
+import '../gs-app';
+import { withComponentDocs } from '../../../.storybook/ComponentDocsBlock';
+import { LAPIS_URL } from '../../constants';
+import type { GenomeDataViewerProps } from '../../preact/genomeViewer/genome-data-viewer';
+import { withinShadowRoot } from '../withinShadowRoot.story';
+
+const codeExample = String.raw`
+<gs-genome-data-viewer
+    gff3Source='https://raw.githubusercontent.com/nextstrain/nextclade_data/8f2e791d3a59013ee88e1d1d7e83b486d39c4ecb/data/nextstrain/wnv/all-lineages/genome_annotation.gff3'
+    genomeLength=11029
```
fengelniederhammer, line 15: I believe that you always need quotes in HTML, don't you?
```suggestion
    genomeLength="11029"
```

components/src/web-components/visualization/gs-genome-data-viewer.stories.ts
```diff
@@ -0,0 +1,112 @@
...
+                    },
+                },
+            ],
+        },
+    },
+    play: async ({ canvasElement, step }) => {
+        const canvas = await withinShadowRoot(canvasElement, 'gs-genome-data-viewer');
+
+        await step('CDS visible', async () => {
+            const all = await canvas.findAllByText('NS1');
```
fengelniederhammer, line 105: This could use a better name, e.g.
```suggestion
            const ns1NameMatches = await canvas.findAllByText('NS1');
```

components/src/preact/genomeViewer/loadGff3.tsx
```diff
@@ -0,0 +1,171 @@
...
+export function loadGenomeLength(content: string) {
+    const lines = content.split('\n');
+    for (const line of lines) {
+        if (!line.startsWith('#')) {
+            continue;
+        }
+        const fields = line.split(' ');
+        if (fields[0] === '##sequence-region') {
+            const start = fields[fields.length - 2];
+            const end = fields[fields.length - 1];
```
fengelniederhammer, line 162: Nitpick:
```suggestion
            const start = fields.at(-2);
            const end = fields.at(-1);
```

components/src/preact/genomeViewer/loadGff3.tsx
fengelniederhammer, line 1: This could be a `.ts` file.

components/src/preact/genomeViewer/loadGff3.tsx
```diff
@@ -0,0 +1,171 @@
+import { UserFacingError } from '../components/error-display';
+import { ColorsRGB, type GraphColor } from '../shared/charts/colors';
+
+export type CDSFeature = {
+    positions: position[];
+    label: string;
+    color: GraphColor;
+};
+
+type position = {
```
fengelniederhammer, line 10:
```suggestion
type Position = {
```

components/src/preact/genomeViewer/loadGff3.tsx
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+        if (!added) {
+            nonOverlappingCDSFeatures.push([cdsFeature]);
+        }
+    }
+    return nonOverlappingCDSFeatures;
+}
+
+export function loadGenomeLength(content: string) {
```
fengelniederhammer, line 153: This function is not covered by a unit test.

components/src/preact/genomeViewer/loadGff3.tsx
```diff
@@ -0,0 +1,171 @@
...
+    color: GraphColor;
+};
+
+type position = {
+    start: number;
+    end: number;
+};
+
+type CDSMap = {
+    [id: string]: { positions: position[]; label: string };
```
fengelniederhammer, line 16: Out of curiosity: Why does this actually have multiple positions? Is it possible that a gene is split into several regions on the genome?

fengelniederhammer (commented):
Looks quite good, just a couple of things :+1:

components/src/web-components/visualization/gs-genome-data-viewer.tsx
```diff
@@ -0,0 +1,74 @@
...
+    interface HTMLElementTagNameMap {
+        'gs-genome-data-viewer': GenomeDataViewerComponent;
+    }
+}
+
+declare global {
+    // eslint-disable-next-line @typescript-eslint/no-namespace
+    namespace JSX {
+        interface IntrinsicElements {
+            'gs-sequences-by-location': DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>;
```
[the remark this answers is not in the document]
anna-parker: wupps! thanks!

components/src/preact/genomeViewer/loadGff3.tsx
anna-parker: yes this happens sometimes (if you are curious you can look into splicing, that is one example of when this happens :-) )

gordonkoehn:
Great feature to see coming !

components/src/preact/genomeViewer/CDSPlot.tsx
```diff
@@ -0,0 +1,219 @@
...
+function getMaxTickNumber(fullWidth: number): number {
+    const baseValue = 1000;
+
+    const normalizedRatio = fullWidth / baseValue;
+
+    const ticks = Math.round(MAX_TICK_NUMBER * normalizedRatio);
+    return Math.max(MIN_TICK_NUMBER, Math.min(MAX_TICK_NUMBER, ticks));
+}
+
+function getTicks(zoomStart: number, zoomEnd: number, fullWidth: number) {
```
fengelniederhammer, line 21: Actually I think that this function could also use a unit test, but I don't want to block too much on this.

fengelniederhammer (approved):

components/src/preact/genomeViewer/CDSPlot.tsx
anna-parker: I will merge then - we will anyways need to probably iterate on this function the more examples we have :-)

PR #816, opened 2025-03-17
fix(components): gs-date-range-filter make it explicit that the event contains `null` when deleting the input

fengelniederhammer:
Discovered in the context of https://github.com/GenSpectrum/dashboards/issues/617

### Summary

The code seems to have sent `undefined` with the event "gs-date-range-option-changed", but both Chrome and Firefox show `null` as event.detail.
### PR Checklist


JonasKellerer (approved):
LGTM

PR #820, opened 2025-03-18
fix(components): disable outline selecting map regions in Chrome

### LINKED ISSUE #729: Map component: Get rid of weird bounding box
(On chrome) Now when clicking on countries in the map component a weird bounding box is added, the box can't handle countries that span larger sections of the globe and weirdly cover some - yet not all countries - with the box outline (see in example Greenland is on top of the box but most African countries are covered). In my opinion this adds nothing to the plot, looks bad and it quite confusing. I think it should be removed:

Image

pflanze:
Chrome has a default CSS enty that enables "bounding boxes" around SVG paths when selecting them; Firefox does not show those (by default). Explicitly disable them as they are distracting.

resolves #729

### PR Checklist

### DIFF
--- a/components/src/preact/sequencesByLocation/leafletStyleModifications.css
+++ b/components/src/preact/sequencesByLocation/leafletStyleModifications.css
@@ -1,3 +1,8 @@
 .leaflet-container {
     background: transparent;
 }
+
+.leaflet-interactive {
+    outline: none;
+    cursor: pointer;
+}


pflanze:
Remember that you can only reproduce it in Chrome or Chromium. Also, it looks like "npm run build" is necessary to actually update the file being sent.

fengelniederhammer (approved):
Good find :)

Issue #792, opened 2025-02-25
Mutation filter: Validate mutations


anna-parker:
![Image](https://github.com/user-attachments/assets/f8652942-233f-469d-8cc1-ccb1a849f573)

PR #826, opened 2025-03-24
feat(components): validate mutations in gs-mutation-filter

JonasKellerer:
Resolves: #792

### Summary

Changes:
- uses downshift for multi select
- changed design, so new mutations are on the same line as the input
- moves the info button to the end of the input
- validates the input when writing, only provides valid options to select
- validates mutation position
- parses pasted comma separated list of mutations and only uses valid mutations
- checks if input is empty on blur, else shows error boundary

### Screenshot

With open menu
grafik

With not submitted mutation:
grafik

### PR Checklist


components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -24,55 +25,63 @@ const mutationFilterPropsSchema = mutationFilterInnerPropsSchema.extend({
...
             <div style={width}>
-                <MutationFilterInner initialValue={initialValue} />
+                <MutationFilterNewInner initialValue={initialValue} />
             </div>
         </ErrorBoundary>
     );
 };
 
-export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> = ({ initialValue }) => {
+function MutationFilterNewInner({ initialValue }: MutationFilterInnerProps) {
```
fengelniederhammer, line 70:
```suggestion
function MutationFilterInner({ initialValue }: MutationFilterInnerProps) {
```
:)

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -24,55 +25,63 @@ const mutationFilterPropsSchema = mutationFilterInnerPropsSchema.extend({
...
-    }
-
-    const handleRemoveValue = (option: ParsedMutationFilter) => {
-        const newSelectedFilters = {
-            ...selectedFilters,
-            [option.type]: selectedFilters[option.type].filter((i) => option.value.toString() != i.toString()),
-        };
+    const items = useMemo(() => {
+        return itemCandidate ? [itemCandidate] : [];
+    }, [itemCandidate]);
```
fengelniederhammer, line 82: That doesn't necessarily need a useMemo, the computation inside is quite trivial.

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -262,82 +261,69 @@ function getPlaceholder(referenceGenome: ReferenceGenome) {
...
+            aminoAcidMutations: [],
+            nucleotideMutations: [],
+            aminoAcidInsertions: [],
+            nucleotideInsertions: [],
+        } as {
+            aminoAcidMutations: string[];
+            nucleotideMutations: string[];
+            aminoAcidInsertions: string[];
+            nucleotideInsertions: string[];
+        },
```
fengelniederhammer, line 327:
```suggestion
        {
            aminoAcidMutations: [],
            nucleotideMutations: [],
            aminoAcidInsertions: [],
            nucleotideInsertions: [],
        } as MutationsFilter,
```

We could even get rid of the `as` by adding the type either as return type of the method or as generic argument to `reduce`.

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -24,55 +25,63 @@ const mutationFilterPropsSchema = mutationFilterInnerPropsSchema.extend({
...
     const { width, initialValue } = props;
     return (
         <ErrorBoundary
-            size={{ height: '3.375rem', width }}
+            size={{ height: '40px', width }}
             layout='horizontal'
             schema={mutationFilterPropsSchema}
             componentProps={props}
         >
             <div style={width}>
```
fengelniederhammer, line 63: Huh, that already didn't work before.
```suggestion
            <div style={{width}}>
```

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -84,170 +93,160 @@ export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> =
...
+            setInputValue(invalidInput);
+            setItemCandidate(null);
+        } else {
+            setInputValue(newInputValue ?? '');
+            if (newInputValue !== undefined) {
+                const candidate = parseAndValidateMutation(newInputValue, referenceGenome);
+                if (candidate) {
+                    setItemCandidate(candidate);
+                } else {
+                    setItemCandidate(null);
```
fengelniederhammer, line 132: `candidate` is already a mutation or null.
```suggestion
                    setItemCandidate(candidate);
```

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -84,170 +93,160 @@ export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> =
...
+    const values = Array.isArray(initialValue) ? initialValue : Object.values(initialValue).flatMap((it) => it);
+
+    return values.reduce((selectedItems, value) => {
+        const parsedMutation = parseAndValidateMutation(value, referenceGenome);
+        if (parsedMutation === null) {
+            return selectedItems;
+        }
+
+        return [...selectedItems, parsedMutation];
+    }, [] as MutationFilterItem[]);
```
fengelniederhammer, line 248:
```suggestion
return values
    .map(value => parseAndValidateMutation(value, referenceGenome))
    .filter((parsedMutation) => parsedMutation !== null);
```

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -24,55 +25,63 @@ const mutationFilterPropsSchema = mutationFilterInnerPropsSchema.extend({
...
 };
 
-export const MutationFilterInner: FunctionComponent<MutationFilterInnerProps> = ({ initialValue }) => {
+function MutationFilterNewInner({ initialValue }: MutationFilterInnerProps) {
     const referenceGenome = useContext(ReferenceGenomeContext);
-    const [selectedFilters, setSelectedFilters] = useState<SelectedFilters>(
+    const filterRef = useRef<HTMLDivElement>(null);
+    const [inputValue, setInputValue] = useState('');
+    const [selectedItems, setSelectedItems] = useState<MutationFilterItem[]>(
         getInitialState(initialValue, referenceGenome),
```
fengelniederhammer, line 75: I think it might be worth to wrap the `getInitialState` in a `useMemo`.

components/src/preact/mutationFilter/mutation-filter.tsx
fengelniederhammer, line 1: I think we should filter out duplicate mutations. It doesn't make sense to include the same mutation twice. (The previous code does that somehow)

fengelniederhammer (approved):
LGTM

components/src/preact/mutationFilter/mutation-filter.tsx
JonasKellerer: True. New habit from the covariants project. There we have https://www.npmjs.com/package/eslint-plugin-react-perf this eslint rule, which does not allow new arrays and objects inside a react component for performance.

components/src/preact/mutationFilter/mutation-filter.tsx
JonasKellerer: I now added it, that we filter out duplicates.

Issue #809, opened 2025-03-11
mutations over time: prevent mutation cell layout shifts


fengelniederhammer:
When making the screen small, the layout shifts from
![Image](https://github.com/user-attachments/assets/580bd671-6a7d-4cae-9f35-c18e0780c2de)
to
![Image](https://github.com/user-attachments/assets/ecc8d212-64bf-40c2-877c-745401db7d15)

The grey cells should keep their size (stay as high as the colored cells)

PR #827, opened 2025-03-24
fix(components): prevent mutation cell layout shift

JonasKellerer:
Resolves: #809

resolves #

### Summary

The issue was, that the text of the accessiblitiy entry "No data" was wrapped and therefore needed more space.

### PR Checklist


fengelniederhammer (approved):
Tricky. Good find.

PR #828, opened 2025-03-24
fix(components): pin `vite-plugin-dts` to 4.5.0 to export JSX.IntrinsicElements namespace declarations

fengelniederhammer:
### Summary

In the last dependency update (#818), `vite-plugin-dts` was upgraded from 4.5.0 to 4.5.3. After that, our JSX.IntrinsicElements declarations (e.g. this one: https://github.com/GenSpectrum/dashboard-components/blob/0c7d6cab120e59295db70495b4467b4c1e463dae/components/src/web-components/gs-app.ts#L110-L117) were not included in the built output anymore (in `dist/components.d.ts`). Thus, the typecheck in the dashboards failed, claiming that e.g. `gs-app` is not a valid JSX component name. 4.5.0 is the last version that works.

https://github.com/qmhc/vite-plugin-dts/issues/419 seems to be the exact same issue.

### PR Checklist

### DIFF
--- a/components/package-lock.json
+++ b/components/package-lock.json
@@ -74,9 +74,9 @@
                 "storybook": "^8.0.9",
                 "storybook-addon-fetch-mock": "^2.0.0",
                 "tailwindcss": "^4.0.9",
-                "typescript": "~5.8.2",
+                "typescript": "^5.8.2",
                 "vite": "^6.0.3",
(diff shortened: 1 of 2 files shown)


anna-parker (approved):
lgtm - thanks so much for looking into this @fengelniederhammer - I can't test well on mac but it makes a lot of sense for this to be the culprit!

PR #840, opened 2025-03-31
feat(components): gs-wastewater-mutations-over-time: persist pagination page size across tabs

### LINKED ISSUE #835: Persist pagination state for gs-wastewater-mutations-over-time between tabs changes
When using the component and switching tabs, it would be nice that the pagination state would persist between tabs. Currently it switches back to the default.

fengelniederhammer:
resolves #835

### PR Checklist


components/src/preact/shared/tanstackTable/pagination-context.tsx
```diff
@@ -0,0 +1,28 @@
+import { createContext, type FunctionComponent } from 'preact';
+import { type Dispatch, type StateUpdater, useContext, useState } from 'preact/hooks';
+
+import type { PageSizes } from './pagination';
+
+type PageSizeContext = {
+    pageSize: number;
+    setPageSize: Dispatch<StateUpdater<number>>;
```
JonasKellerer, line 8: Didn't know this type exists. Nice. We could use that at some other point/PR also in the `Toolbarprops`.

components/src/preact/shared/tanstackTable/pagination.tsx
```diff
@@ -59,9 +65,11 @@ function PageSizeSelector({
             <div className={'text-nowrap text-sm'}>Rows per page:</div>
             <select
                 className={`select select-ghost select-sm ${heightForSmallerLines}`}
-                value={table.getState().pagination.pageSize}
+                value={pageSize}
                 onChange={(e) => {
-                    table.setPageSize(Number(e.currentTarget?.value));
+                    const pageSize = Number(e.currentTarget?.value);
```
JonasKellerer, line 70: Should we add a default here? I think `Number(undefined)` is `NaN` and it would be strange to set it to that value.

components/src/preact/shared/tanstackTable/pagination.tsx
```diff
@@ -90,7 +98,7 @@ function GotoPageSelector({ table }: PaginationProps) {
                 defaultValue={table.getState().pagination.pageIndex + 1}
                 onChange={(e) => {
                     const page = e.currentTarget.value ? Number(e.currentTarget.value) - 1 : 0;
-                    table.setPageIndex(page);
+                    table.setPageIndex(Math.min(page, table.getPageCount() - 1));
```
JonasKellerer, line 101: Nice. I have only fixed that in covariants and not here 🙈

components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.stories.tsx
```diff
@@ -67,6 +67,22 @@ export const Default: StoryObj<WastewaterMutationsOverTimeProps> = {
...
+        await step('Wait for component to render', async () => {
+            await canvas.findByText('Lugano');
+        });
+
+        const getRowsPerPageSelectors = async () => await canvas.findAllByLabelText('Rows per page', { exact: false });
+
+        await expect((await getRowsPerPageSelectors())[0]).toHaveValue('10');
+        await expect((await getRowsPerPageSelectors())[1]).toHaveValue('10');
+        await userEvent.selectOptions((await getRowsPerPageSelectors())[0], '20');
+        await expect((await getRowsPerPageSelectors())[1]).toHaveValue('20');
```
JonasKellerer, line 82: Nitpick: In the other stories we have put the test also in a step, so we can provide a description to it. Here we also have it in the name of the story so I am fine with this.

components/src/preact/shared/tanstackTable/tanstackTable.tsx
```diff
@@ -39,5 +45,13 @@ export function usePreactTable<TData extends RowData>(options: TableOptions<TDat
...
     }));
 
+    const { pageSize } = usePageSizeContext();
+    useEffect(
+        () => {
+            tableRef.current.setPageSize(pageSize);
+        },
+        [pageSize], // eslint-disable-line react-hooks/exhaustive-deps -- only run this when the pageSize changes
+    );
+
```
JonasKellerer, line 55: This way we can no longer use `usePreactTable` without the context. Do we want that? I could imagine that we need a table at some point, that does not have pagination.

JonasKellerer: Could we place this also somewhere else?

JonasKellerer (commented):
One open qustion regarding the usePageSizeContext in the tanstackTable.tsx. Appart from that it looks nice.

components/src/preact/wastewater/mutationsOverTime/wastewater-mutations-over-time.stories.tsx
fengelniederhammer: Steps have the disadvantage that you can only rerun the whole step as a whole, but not the individual things inside a step.

But I'll change it.

components/src/preact/shared/tanstackTable/tanstackTable.tsx
fengelniederhammer: I believe you could still use it without the context. The context values will be useless, because it only has its default values, but that's fine if you don't need pagination I guess. -> I'd prefer to leave it here.

components/src/preact/shared/tanstackTable/pagination-context.tsx
fengelniederhammer: The return type of `useState` tells you why type the setter has ;)

components/src/preact/shared/tanstackTable/pagination.tsx
fengelniederhammer: I can't really imagine how we could get into a state that leaves us with `NaN` here? I also don't know what a reasonable default should be -> I'll throw instead.

components/src/preact/shared/tanstackTable/tanstackTable.tsx
JonasKellerer: I tried it out. It still works. However, I dont like that we couple the context with the table, introducing a side-effect.
Maybe we could do something like this, to make it explicit. 
```
export function useWithPaginationContext<TData extends RowData>(table: Table<TData>) {
    const { pageSize } = usePageSizeContext();
    useEffect(
        () => {
            table.setPageSize(pageSize);
        },
        [pageSize], // eslint-disable-line react-hooks/exhaustive-deps -- only run this when the pageSize changes
    );
    return table;
}
```
and use it like this:

```
const table = useWithPaginationContext(
        usePreactTable({
            data: tableData,
            columns,
            getCoreRowModel: getCoreRowModel(),
        }),
    );
```

PR #843, opened 2025-04-01
feat(components): gs-mutations-over-time search for mutations and annotations

### LINKED ISSUE #805: mutations over times: search for mutations
Add a search field that does full text search on all mutations in the list. This should match mutation annotations.

JonasKellerer:
Resolves #805

resolves #

### Summary

Adds a mutation filter in the toolbar. It is possible to filter for mutations and for content of the annotations. 

In a second step, we could add checkboxes to the tooltip, so users can select the annotations they want to see. Currently, they have to type in content from the annotation or use the symbol.

### PR Checklist


fengelniederhammer (approved):
LGTM

PR #849, opened 2025-04-08
feat(components): render calendar inside shadow dom

### LINKED ISSUE #806: Render date pickers in the shadow DOM
flatpickr (the date pickers of gs-date-range-filter) renders itself into the light DOM and thus needs the corresponding styles to be available in the light DOM. -> find out whether we can also render that into the shadow DOM

JonasKellerer:
Resolves #806

### Summary

Flatpickr has a built in method to render the calendar to other elements. I just used that. This way the calendar is inside the shadow dom and we can move the css of flatpickr inside our components.

We can see if we can use this also for our other dropdown components.

### PR Checklist

### DIFF
--- a/components/src/preact/dateRangeFilter/date-picker.tsx
+++ b/components/src/preact/dateRangeFilter/date-picker.tsx
@@ -22,8 +22,10 @@ export function DatePicker({
 
     const [datePicker, setDatePicker] = useState<flatpickr.Instance | null>(null);
 
+    const calendarRef = useRef<HTMLDivElement>(null);
+
     useEffect(() => {
-        if (!inputRef.current) {
+        if (!inputRef.current || !calendarRef.current) {
             return;
         }
 
@@ -33,6 +35,7 @@ export function DatePicker({
             defaultDate: value,
             minDate,
             maxDate,
+            appendTo: calendarRef.current,
         });
 
         setDatePicker(instance);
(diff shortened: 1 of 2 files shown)


components/src/web-components/PreactLitAdapterWithFlatpickrStyles.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { unsafeCSS } from 'lit';
+
+import { PreactLitAdapter } from './PreactLitAdapter';
+
+import 'flatpickr/dist/flatpickr.css';
+
+const flatpickrCss = unsafeCSS(flatpickrStyle);
+
+export abstract class PreactLitAdapterWithFlatpickrStyles extends PreactLitAdapter {
+    static override styles = [...PreactLitAdapter.styles, flatpickrCss];
```
fengelniederhammer, line 11: We'll likely only need those styles in one component (i.e. in the date range filter). IMO it would also be ok to delete this class and move this line to the DateRangeFilterComponent instead.

components/src/web-components/PreactLitAdapterWithFlatpickrStyles.tsx
```diff
@@ -0,0 +1,12 @@
+import flatpickrStyle from 'flatpickr/dist/flatpickr.css?inline';
+import { unsafeCSS } from 'lit';
+
+import { PreactLitAdapter } from './PreactLitAdapter';
+
+import 'flatpickr/dist/flatpickr.css';
```
fengelniederhammer, line 6: We probably don't need this import, do we?

fengelniederhammer (approved):
LGTM

components/src/web-components/PreactLitAdapterWithFlatpickrStyles.tsx
JonasKellerer: I now moved it to `gs-date-range-filter.tsx` and removed this class.

components/src/web-components/PreactLitAdapterWithFlatpickrStyles.tsx
JonasKellerer: True.

PR #850, opened 2025-04-09
fix(components): apply daisyui styles to host instead of root

### LINKED ISSUE #804: Get rid of `styles.css`
We create a `styles.css` when building the component that should actually be unnecessary. However, you need to include it when using the compents. This is technical debt that spills our styles into the surrounding page.

It is necessary due to two problems:
* #806 
* #807

### LINKED ISSUE #807: Make sure that DaisyUI has all styles available in the shadow DOM without any additional styles necessary
DaisyUI needs some CSS properties to be set in the surrounding page. Find out why - actually it should suffice to have them set to the shadow root (That's how some of the Tailwind properties work). Try to isolate the problem to understand why it doesn't work.

fengelniederhammer:
resolves #807
resolves #804

### Summary

See https://daisyui.com/docs/config/#root
The default is
```
@plugin "daisyui" {
  root: ':root';
}
```
which only targets the HTML document. We need to apply the styles to `:host`.

### PR Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,3 +29,18 @@ where
 * `<name of the upstream branch>` is the name of the branch for which the changelog should be generated.
 
 __NOTE: This command does not respect local changes. It will pull the commit messages from the remote repository.__
+
+## CSS
+
+We use [Tailwind](https://tailwindcss.com/) and [DaisyUI](https://daisyui.com/) for styling.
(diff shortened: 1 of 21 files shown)


fengelniederhammer:
For the record: It's the same problem as https://github.com/tailwindlabs/tailwindcss/discussions/16772.

Tailwind uses `@property` which doesn't work inside the shadow DOM.

PR #852, opened 2025-04-13
feat(components): gs-text-input: make it possible to paste values and accept them on blur

### LINKED ISSUE #812: Text filter: Copying a value into the filter should update the value
Current behavior: 
* Copy a valid value into the filter
* The value is not updated, no event is thrown
  * IMO we should already store the value at this point and throw an event.
* One needs to "confirm" the value by clicking on it in the autocomplete dropdown.

fengelniederhammer:
resolves #812

### Summary

We have to control the `selectedItem` ourselves instead of leaving it to downshift, because otherwise it will be reset to the `value` that is provided to the web component.

### PR Checklist

### DIFF
--- a/components/src/preact/components/downshift-combobox.tsx
+++ b/components/src/preact/components/downshift-combobox.tsx
@@ -1,6 +1,6 @@
 import { useCombobox } from 'downshift/preact';
 import { type ComponentChild } from 'preact';
-import { useMemo, useRef, useState } from 'preact/hooks';
+import { useEffect, useMemo, useRef, useState } from 'preact/hooks';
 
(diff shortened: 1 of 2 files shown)


components/src/preact/components/downshift-combobox.tsx
```diff
@@ -67,17 +75,25 @@ export function DownshiftCombobox<Item>({
         itemToString(item) {
             return itemToString(item);
         },
-        selectedItem: value,
```
anna-parker, line 70: why is this nolonger required?

anna-parker (commented):
lgtm

PR #855, opened 2025-04-15
761 number input draft a suggestion

### LINKED ISSUE #761: Number input: draft a suggestion

fengelniederhammer:
resolves #761

### PR Checklist

### DIFF
--- a/components/src/web-components/input/gs-text-filter.stories.ts
+++ b/components/src/web-components/input/gs-text-filter.stories.ts
@@ -18,8 +18,8 @@ const codeExample = String.raw`
     lapisFilter='{"country": "Germany"}'
     placeholderText="Enter host name"
     value="Homo sapiens"
-    width="50%">
-</gs-text-filter>`;
+    width="50%"
+></gs-text-filter>`;
 
 const meta: Meta<Required<TextFilterProps>> = {
     title: 'Input/Text filter',
--- a/components/src/preact/components/min-max-range-slider.tsx
+++ b/components/src/preact/components/min-max-range-slider.tsx
@@ -6,6 +6,7 @@ export interface MinMaxPercentSliderProps {
     max: number;
     setMin: (min: number) => void;
     setMax: (max: number) => void;
+    onDrop?: () => void;
     rangeMin?: number;
     rangeMax?: number;
     step?: number;
@@ -16,6 +17,7 @@ export const MinMaxRangeSlider: FunctionComponent<MinMaxPercentSliderProps> = ({
     max,
     setMin,
     setMax,
+    onDrop,
     rangeMin = 0,
     rangeMax = 100,
     step = 0.1,
@@ -55,14 +57,16 @@ export const MinMaxRangeSlider: FunctionComponent<MinMaxPercentSliderProps> = ({
         }
     };
 
+    const lowerBoundary = getGradientBoundary(min, rangeMin, rangeMax);
+    const upperBoundary = getGradientBoundary(max, rangeMin, rangeMax);
     const background = `
         linear-gradient(
             to right,
             ${sliderColor} 0%,
-            ${sliderColor} ${min}%,
-            ${rangeColor} ${min}%,
-            ${rangeColor} ${max}%,
-            ${sliderColor} ${max}%,
+            ${sliderColor} ${lowerBoundary}%,
+            ${rangeColor} ${lowerBoundary}%,
+            ${rangeColor} ${upperBoundary}%,
+            ${sliderColor} ${upperBoundary}%,
             ${sliderColor} 100%)
     `;
 
@@ -73,6 +77,8 @@ export const MinMaxRangeSlider: FunctionComponent<MinMaxPercentSliderProps> = ({
                 type='range'
                 value={min}
                 onInput={onMinChange}
+                onMouseUp={() => onDrop?.()}
+                onTouchEnd={() => onDrop?.()}
                 min={`${rangeMin}`}
                 max={`${rangeMax}`}
                 step={step}
@@ -86,8 +92,17 @@ export const MinMaxRangeSlider: FunctionComponent<MinMaxPercentSliderProps> = ({
                 max={`${rangeMax}`}
                 step={step}
                 onInput={onMaxChange}
+                onMouseUp={() => onDrop?.()}
+                onTouchEnd={() => onDrop?.()}
                 style={{ background, zIndex: zIndexTo }}
             />
         </div>
     );
 };
+
+/**
+ * This is a linear function that returns 0 for x = lowerBound and 100 for x = upperBound.
+ */
+function getGradientBoundary(x: number, lowerBound: number, upperBound: number) {
+    return ((x - lowerBound) / (upperBound - lowerBound)) * 100;
+}
--- a/components/src/preact/dateRangeFilter/date-range-filter.stories.tsx
+++ b/components/src/preact/dateRangeFilter/date-range-filter.stories.tsx
@@ -256,7 +256,10 @@ export const ChangingTheValueProgrammatically: StoryObj<DateRangeFilterProps> = 
 
         await step('Clearing the value from within the component is still possible', async () => {
             await waitFor(async () => {
-                await userEvent.click(canvas.getByRole('button', { name: '×' }));
+                await expect(canvas.getByRole('button', { name: '×' })).toBeVisible();
+            });
+            await userEvent.click(canvas.getByRole('button', { name: '×' }));
+            await waitFor(async () => {
                 await expectOptionSelected(canvasElement, placeholder);
             });
             await waitFor(async () => {
--- a/components/src/preact/numberRangeFilter/NumberRangeFilterChangedEvent.ts
+++ b/components/src/preact/numberRangeFilter/NumberRangeFilterChangedEvent.ts
@@ -1,0 +1,31 @@
+import z from 'zod';
+
+import { gsEventNames } from '../../utils/gsEventNames';
+
+type LapisNumberFilter = Record<string, number | undefined>;
+
+export const numberRangeSchema = z.object({
+    min: z.number().optional(),
+    max: z.number().optional(),
+});
+export type NumberRange = z.infer<typeof numberRangeSchema>;
+
+export class NumberRangeValueChangedEvent extends CustomEvent<NumberRange> {
+    constructor(detail: NumberRange) {
+        super(gsEventNames.numberRangeValueChanged, {
+            detail,
+            bubbles: true,
+            composed: true,
+        });
+    }
+}
+
+export class NumberRangeFilterChangedEvent extends CustomEvent<LapisNumberFilter> {
+    constructor(detail: LapisNumberFilter) {
+        super(gsEventNames.numberRangeFilterChanged, {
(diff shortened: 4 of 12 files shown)


fengelniederhammer:
TODOs:
* fix layout
image
* test on mobile
* test on dashboards
* write docs

components/src/preact/numberFilter/number-filter.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { NumberFilterChangedEvent } from './NumberFilterChangedEvent';
+import { type NumberRange, SetRangeActionType, useSelectedRangeReducer } from './useSelectedRangeReducer';
+import { ErrorBoundary } from '../components/error-boundary';
+import { UserFacingError } from '../components/error-display';
+import { MinMaxRangeSlider } from '../components/min-max-range-slider';
+import { ResizeContainer } from '../components/resize-container';
+import { DeleteIcon } from '../shared/icons/DeleteIcon';
+
+const numberFilterPropsSchema = z.object({
+    value: z.record(z.number().optional()),
```
anna-parker, line 14: so this is actually a valueMap right?

anna-parker: it is a map that must contain ${lapisField}From and ${lapisField}To

components/src/preact/numberFilter/useSelectedRangeReducer.ts
```diff
@@ -0,0 +1,128 @@
...
+    const [range, dispatchRange] = useReducer(
+        rangeReducer,
+        addRange({
+            min: initialValue.min?.toString() ?? '',
+            max: initialValue.max?.toString() ?? '',
+        }),
+    );
+    const [isInitialRender, setIsInitialRender] = useState(true);
+
+    useEffect(
```
anna-parker, line 28: why is this required? I can't see when else this is called apart from the initialRender?

anna-parker (changes requested):
Two little comments :-)

components/src/preact/numberFilter/number-filter.tsx
fengelniederhammer: Yes. But actually I thought about changing it to a fixed object `{ min, max }` and dispatching two events: 
* one that yields the LAPIS filter (i.e. the object `{ ${lapisField}From, ${lapisField}To }`),
* one that can be used as as input (i.e. `{ min, max }`).#

What do you think? We do something similar for the date filter already.

components/src/preact/numberFilter/useSelectedRangeReducer.ts
fengelniederhammer: It will be executed when `initialValue` changes, i.e. it makes it possible to make it a controlled component (controlling the input state in the consuming code).

fengelniederhammer: Removing this `useEffect` should make the `ChangingTheValueProgrammatically` story fail.

components/src/preact/numberFilter/number-filter.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const [currentRange, dispatchRange] = useSelectedRangeReducer(cleanedValue);
+    const [shouldDispatchEvent, setShouldDispatchEvent] = useState(false);
+
+    useEffect(() => {
+        if (!shouldDispatchEvent) {
+            return;
+        }
+
+        setShouldDispatchEvent(false);
+        if (currentRange.range.isValidRange) {
```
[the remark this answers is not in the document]
fengelniederhammer: How bad is that? IMO that would be ok, but we can also decide to only fire events `onBlur`. @chaoran-chen what do you think?

components/src/preact/numberFilter/number-filter.tsx
anna-parker: Oh I do prefer the way it is done in the dateFilter with `dateRangeOptionSchema` - I think it might be nice to do that here as well for consistency (also is clearer when reading through the code :-))

components/src/preact/numberFilter/number-filter.tsx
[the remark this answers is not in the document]
chaoran-chen: Spontaneously, `onBlur` sounds better, for the TextFilter, we also only send an event when an option is chosen from the suggestions or on blur if I'm not mistaken.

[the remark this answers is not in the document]
fengelniederhammer: The difference is that in the text filter you explicitly have to chose one of the available autocomplete options, which is not the case here. In principle we can always send the event.

components/src/preact/numberFilter/number-filter.stories.tsx
```diff
@@ -0,0 +1,377 @@
...
+        await step('make min input larger than max input and check that no event it dispatched', async () => {
+            filterChangedListenerMock.mockClear();
+            valueChangedListenerMock.mockClear();
+            await userEvent.type(minInput(), '8');
+            await changeFocus();
+            await waitFor(async () => {
+                await expect(minInput()).toHaveValue('98');
+                await expect(maxInput()).toHaveValue('90');
+            });
+            await expect(filterChangedListenerMock).not.toHaveBeenCalled();
```
anna-parker, line 127: We could also check that there is an error (turns red)

JonasKellerer:
Just a first thing I noticed: When I enter "-1" in the "from" field, I don't see an error indicator.

JonasKellerer:
When I set the min value of the slider to a negative number, then I see this:

grafik

It seems there is something wrong with the slider component. But I have not checked it.

JonasKellerer:
Just a nitpick: The DateRangeFilter has in the name also "range" since we filter for ranges. Do we want to have the same naming here? So `NumberRangeFilter` or something better?

components/src/preact/numberFilter/number-filter.stories.tsx
fengelniederhammer: How would you do that? Checking colors is difficult AFAIK.

components/src/preact/numberFilter/number-filter.stories.tsx
```diff
@@ -0,0 +1,377 @@
...
+            await waitFor(() => expect(valueChangedListenerMock).toHaveBeenCalled());
+            filterChangedListenerMock.mockClear();
+            valueChangedListenerMock.mockClear();
+        });
+
+        await step('type invalid number into input field', async () => {
+            await userEvent.type(maxInput(), 'not a number');
+            await waitFor(async () => {
+                await expect(minInput()).toHaveValue('10');
+                await expect(maxInput()).toHaveValue('not a number');
```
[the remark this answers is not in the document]
fengelniederhammer: The next step deletes the input (that's why you don't see it anymore). If you execute the test step by step in storybook, you can see it.

components/src/preact/numberFilter/number-filter.stories.tsx
```diff
@@ -0,0 +1,377 @@
...
+        lapisField: '',
+    },
+    play: async ({ canvasElement, step }) => {
+        await step('expect error message', async () => {
+            await expectInvalidAttributesErrorMessage(canvasElement, 'String must contain at least 1 character(s)');
+        });
+    },
+};
+
+export const ChangingTheValueProgrammatically: StoryObj<NumberFilterProps> = {
```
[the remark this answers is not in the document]
fengelniederhammer: We try to split the stories:
* The web component storybook is also public documentation, so we try to keep it basic and stick to stories that are useful to actual users
* The Preact storybook is used for "private" sub-components that are not exposed (buttons, tabs, etc.) and for tests. If we write stories just for the purpose of a unit test, we try to put it in the Preact storybook, if possible. 

For the sake of this test, it doesn't make a difference whether I test the Preact or the web component, so I decided to "hide" it in the Preact storybook.

We discovered that we need this kind of behavior quite late in the process of developing the components. It would be good if every input component had this kind of test, but I think we didn't get to add it for all of them.

components/src/web-components/input/gs-number-filter.tsx
```diff
@@ -0,0 +1,147 @@
...
+    @property({ type: Number })
+    sliderMax = 100;
+
+    /**
+     * The `step` value to use for the slider.
+     *
+     * This attribute has no effect on the text input.
+     */
+    @property({ type: Number })
+    sliderStep = 0.1;
```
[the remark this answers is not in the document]
fengelniederhammer: Good point, I'll change it.

components/src/web-components/input/gs-number-filter.tsx
```diff
@@ -0,0 +1,147 @@
...
+ * ## Context
+ *
+ * This component lets the user specify filters for a `float` or `int` field in LAPIS.
+ * It consists of two text input fields and a slider.
+ *
+ * The slider must be restricted to a certain range.
+ * Users can however still enter values outside of this range in the text input fields.
+ *
+ * @fires {CustomEvent<Record<string, string | undefined>>} gs-number-filter-changed
+ * Fired when the slider is released, when the user types a valid range in the input fields,
```
JonasKellerer, line 23: This is only partially correct, since we only fire the event on blur, and not, when the user types. But I am not sure if we need this in this docs?

components/src/web-components/input/gs-text-filter.stories.ts
```diff
@@ -18,8 +18,8 @@ const codeExample = String.raw`
     lapisFilter='{"country": "Germany"}'
     placeholderText="Enter host name"
     value="Homo sapiens"
-    width="50%">
-</gs-text-filter>`;
+    width="50%"
+></gs-text-filter>`;
```
JonasKellerer, line 22: Good find.

JonasKellerer (approved):
Looks good to me. 

Did you test it in the dashboards, meaning, does the design fit into our other filters?

components/src/web-components/input/gs-number-filter.tsx
fengelniederhammer: Good catch, I forgot to change the docs after I changed the code...

components/src/preact/numberFilter/number-filter.stories.tsx
fengelniederhammer: I added it to the `<input>` elements, the test seems to work.

components/src/preact/numberFilter/number-filter.tsx
```diff
@@ -0,0 +1,153 @@
...
+        scheduleEventDispatch();
+    };
+
+    const inputError = currentRange.range.isValidRange ? '' : 'input-error';
+
+    return (
+        <div ref={divRef}>
+            <div className='join w-full'>
+                <div className={`join-item w-full flex input px-2 ${inputError}`}>
+                    <input
```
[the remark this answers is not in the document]
fengelniederhammer: I split this out to #870

components/src/preact/numberRangeFilter/number-filter.stories.tsx
JonasKellerer, line 1: This file could also be changed to `number-range-filter.stories.tsx`.

JonasKellerer (commented):
I only found one renaming issue :)

PR #873, opened 2025-04-28
feat(components): remove quotes from CDS name

anna-parker:
resolves #

### PR Checklist

### DIFF
--- a/components/src/preact/genomeViewer/loadGff3.ts
+++ b/components/src/preact/genomeViewer/loadGff3.ts
@@ -60,6 +60,13 @@ function getAttributes(attributes: string): Map<string, string> {
     return attrMap;
 }
 
+function removeQuotes(input: string | undefined): string | undefined {
+    if (!input) {
+        return undefined;
+    }
+    return input.replace(/^['"](.*)['"]$/, '$1');
+}
+
 function getCDSMap(lines: string[], genome_type: string, geneMap: CDSMap): CDSMap {
     for (const line of lines) {
         if (line.startsWith('#') || line.trim() === '') {
@@ -73,7 +80,7 @@ function getCDSMap(lines: string[], genome_type: string, geneMap: CDSMap): CDSMa
 
         const [, , type, startStr, endStr, , , , attributes] = fields;
 
-        if (type.toLowerCase() !== genome_type.toLowerCase()) {
+        if (removeQuotes(type.toLowerCase()) !== genome_type.toLowerCase()) {
             continue;
         }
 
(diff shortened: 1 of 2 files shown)


components/src/preact/genomeViewer/loadGff3.ts
```diff
@@ -60,6 +60,13 @@ function getAttributes(attributes: string): Map<string, string> {
     return attrMap;
 }
 
+function removeQuotes(input: string | undefined): string | undefined {
+    if (!input) {
+        return undefined;
+    }
+    return input.replace(/^['"](.*)['"]$/, '$1');
+}
```
JonasKellerer, line 68: I think this does the same.

```suggestion
function removeQuotes(input: string | undefined) {
    return input?.replace(/^['"](.*)['"]$/, '$1');
}
```

JonasKellerer: I would also just use the input as a string and handle the undefined case before this function. So this would be 
```ts
function removeQuotes(input: string) {
    return input.replace(/^['"](.*)['"]$/, '$1');
}
```

components/src/preact/genomeViewer/loadGff3.ts
```diff
@@ -85,16 +92,16 @@ function getCDSMap(lines: string[], genome_type: string, geneMap: CDSMap): CDSMa
         }
 
         const attrPairs = getAttributes(attributes);
-        const label = attrPairs.get('Name') || attrPairs.get('gene') || attrPairs.get('gene_name');
+        const label = removeQuotes(attrPairs.get('Name') || attrPairs.get('gene') || attrPairs.get('gene_name'));
```
JonasKellerer, line 95: After the comment above, this would now be:
```ts
const labelAttribute = attrPairs.get('Name') || attrPairs.get('gene') || attrPairs.get('gene_name');
if (!labelAttribute ) {
            throw new UserFacingError(
                'Invalid gff3 source',
                `No label found for feature: "${line}", must contain label in Name, gene, or gene_name attribute`,
            );
        }
  const label = removeQuotes(labelAttribute);
```
We then could get rid of the `!` below.

components/src/preact/genomeViewer/loadGff3.ts
```diff
@@ -85,16 +92,16 @@ function getCDSMap(lines: string[], genome_type: string, geneMap: CDSMap): CDSMa
...
             throw new UserFacingError(
                 'Invalid gff3 source',
                 `No label found for feature: "${line}", must contain label in Name, gene, or gene_name attribute`,
             );
         }
-        const id = attrPairs.get('ID') || label;
+        const id = removeQuotes(attrPairs.get('ID') || label)!;
         if (attrPairs.get('Parent')) {
-            const parent = attrPairs.get('Parent');
+            const parent = removeQuotes(attrPairs.get('Parent'));
```
JonasKellerer, line 104: And this would then be:
```ts
const parentAttribute = attrPairs.get('Parent');
        if (parentAttribute) {
            const parent = removeQuotes(parentAttribute);
```

JonasKellerer (approved):
Nice. Only some minor changes, so we don't need to disable the type safety.

PR #886, opened 2025-05-12
feat(components): show counts on lineage filter

### LINKED ISSUE #663: Show counts in autocomplete options - gs-lineage-filter

JonasKellerer:
Resolves #663

### Summary

- Loads the lineage definition file, creates a lineage tree of it, and then sums up the values of the aggregated call. 
- As discussed, the component displays values to all lineages, even those with count of 0. The lineages are sorted by their name, not the count to make it less confusing for the user.
- There is some potential for optimization, but currently it seems this is not necessary

### PR Checklist


components/src/preact/lineageFilter/lineage-filter.tsx
```diff
@@ -64,30 +65,41 @@ const LineageFilterInner: FunctionComponent<LineageFilterInnerProps> = ({
     return <LineageSelector lapisField={lapisField} value={value} placeholderText={placeholderText} data={data} />;
 };
 
+type LineageItem = { lineage: string; count: number };
```
fengelniederhammer, line 68: Let's more this type to `fetchLineageAutocompleteList.ts`, because that's where the objects are created? That would be more self-consistent.

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
```diff
@@ -1,20 +1,116 @@
...
-    const fetchAggregatedOperator = new FetchAggregatedOperator<Record<string, string>>(lapisFilter ?? {}, [field]);
+    const fetchAggregatedOperator = new FetchAggregatedOperator<Record<string, string>>(lapisFilter ?? {}, [
+        lapisField,
+    ]);
+
+    const countsByLineageArray = (await fetchAggregatedOperator.evaluate(lapisUrl, signal)).content;
+    return countsByLineageArray.reduce(
+        (acc, value) => acc.set(value[lapisField], value.count),
+        new Map<string, number>(),
+    );
```
fengelniederhammer, line 62: I think this does the same:
```suggestion
    return new Map<string, number>(countsByLineageArray.map((value) => [value[lapisField], value.count]));
```

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
```diff
@@ -1,20 +1,116 @@
...
+    lapisUrl: string;
+    lapisField: string;
+    signal?: AbortSignal;
+}) {
+    const lineageDefinitions = await fetchLineageDefinition({ lapisUrl, lapisField, signal });
+
+    const lineageTree = new Map<string, { children: string[] }>();
+
+    Object.entries(lineageDefinitions).forEach(([lineage, definition]) => {
+        lineageTree.set(lineage, { children: [] });
```
fengelniederhammer, line 79: I think this line makes the parent forget its children in case the `forEach` below was executed first. I'll comment separately on a test that would catch that.

components/src/preact/lineageFilter/fetchLineageAutocompleteList.spec.ts
```diff
@@ -4,25 +4,283 @@ import { fetchLineageAutocompleteList } from './fetchLineageAutocompleteList';
...
+
+        lapisRequestMocks.lineageDefinition(
+            {
+                A: {
+                    aliases: ['a'],
+                },
+                'A.1': {
+                    parents: ['A'],
+                    aliases: ['a.1'],
+                },
```
fengelniederhammer, line 242: Let's mix up the order a little to make sure that we're indifferent to how the file was sorted.
```suggestion
                'A.1': {
                    parents: ['A'],
                    aliases: ['a.1'],
                },
                A: {
                    aliases: ['a'],
                },
```

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
```diff
@@ -1,20 +1,116 @@
...
+                    lineage,
+                    count: countsByLineage.get(lineage) ?? 0,
+                },
+                {
+                    lineage: `${lineage}*`,
+                    count: getCountsIncludingSublineages(lineage, lineageTree, countsByLineage),
+                },
+            ];
+        })
+        .sort((a, b) => a[0].lineage.localeCompare(b[0].lineage))
```
fengelniederhammer, line 39: We can sort first. That seems a little easier.
```suggestion
        .sort((a, b) => a.localeCompare(b))
        .map((lineage) => {
            return [
                {
                    lineage,
                    count: countsByLineage.get(lineage) ?? 0,
                },
                {
                    lineage: `${lineage}*`,
                    count: getCountsIncludingSublineages(lineage, lineageTree, countsByLineage),
                },
            ];
        })
```

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
```diff
@@ -1,20 +1,116 @@
...
+
+    const countOfChildren = [...descendents].reduce((sum, child) => {
+        return sum + (countsByLineage.get(child) ?? 0);
+    }, 0);
+    const countLineage = countsByLineage.get(lineage) ?? 0;
+
+    return countOfChildren + countLineage;
+}
+
+function getAllDescendents(lineage: string, lineageTree: Map<string, { children: string[] }>): string[] {
```
fengelniederhammer, line 108: descendAnt seems the more common spelling (rather than descendEnt) according to a quick search.
```suggestion
function getAllDescendants(lineage: string, lineageTree: Map<string, { children: string[] }>): string[] {
```

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
```diff
@@ -1,20 +1,116 @@
...
+
+    return lineageTree;
+}
+
+function getCountsIncludingSublineages(
+    lineage: string,
+    lineageTree: Map<string, { children: string[] }>,
+    countsByLineage: Map<string, number>,
+): number {
+    const descendents = new Set(getAllDescendents(lineage, lineageTree));
```
fengelniederhammer, line 98: Nitpick: We could directly return a set from getAllDescendents

fengelniederhammer (changes requested):
I think I found one bug, otherwise looks good.

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
JonasKellerer: Good thinking.

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
JonasKellerer: I now check first, that the lineage is not yet set, which fixes the error.

JonasKellerer: Good catch btw. :)

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
JonasKellerer: I first decided against it, so the set function has to be called only once. But this makes the return value of the function better.

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
```diff
@@ -1,20 +1,117 @@
...
+    lapisUrl: string;
+    lapisField: string;
+    signal?: AbortSignal;
+}) {
+    const lineageDefinitions = await fetchLineageDefinition({ lapisUrl, lapisField, signal });
+
+    const lineageTree = new Map<string, { children: string[] }>();
+
+    Object.entries(lineageDefinitions).forEach(([lineage, definition]) => {
+        if (lineageTree.get(lineage) === undefined) {
```
fengelniederhammer, line 78:
```suggestion
        if (!lineageTree.has(lineage)) {
```

fengelniederhammer (approved):
LGTM

PR #892, opened 2025-05-15
docs(components): make sure that the storybook docs are comprehensive enough

### LINKED ISSUE #858: Make sure that the Storybook docs are comprehensive enough

fengelniederhammer:
resolves #858

### PR Checklist


components/src/web-components/visualization/introduction.mdx
```diff
@@ -1,10 +1,22 @@
...
+# Introduction
 
-## Scales
+The components in this section visualize data from LAPIS.
+They take a LAPIS filter as input, fetch data from LAPIS and visualize the results in charts and tables.
+
+Refer to the Swagger UI / OpenAPI documentation of your LAPIS instance for the details of what valid LAPIS filters are.
+The LAPIS filter should only contain actual "query filters" (i.e. no `limit`, `orderBy` or `fields`).
+The components perform minor modifications to the LAPIS filter before sending it to LAPIS,
+such as adding stratification by adding `fields`
```
anna-parker, line 13: this is a bit confusing - so when they add `fields` this is not added to the LAPIS filter? Or it is added but `fields` cannot be added to the input LAPIS filter object?

fengelniederhammer: You must not add `fields` yourself. Some (or most?) components will overwrite it. And it would break most components I guess.
Do you have a suggestion what I should change to make it clearer?

components/src/web-components/input/introduction.mdx
```diff
@@ -7,5 +7,61 @@ import { Meta } from '@storybook/blocks';
...
-Every component fires `CustomEvent`s when the user interacts with it, which can be used to update the LAPIS filters.
-The `detail` of the event is designed such that it can be directly passed to the LAPIS API.
+Every component fires `CustomEvent`s when the user interacts with it.
+`event.detail` contains the payload of the event.
+
+Every component fires an event that can be used to update the LAPIS filters.
+It is supposed to be used in the style of:
+
+```javascript
+component.addEventListener('gs-example-event', (event) => {
```
anna-parker, line 17: what about the `componentFinishedLoading` event?

components/src/web-components/input/introduction.mdx
```diff
@@ -7,5 +7,61 @@ import { Meta } from '@storybook/blocks';
...
+In an uncontrolled component, the value is controlled by the DOM and the surrounding Javascript code only reads the value
+(e.g. by listening to events).
+
+All our input components can be used in both ways.
+To make it easier to use the components in a controlled way,
+every component fires an event whose `detail` can be used as the `value` of the component.
+Some components only fire a single event when the event can be used to update the LAPIS filter
+**and** the value of the component.
+Other components fire two different events,
+when the schema of the component’s `value` attribute differs from the schema required for the LAPIS filter.
```
anna-parker, line 38: This is a bit confusing to me, in cases where two events are fired will both include a details section and how will the user know which can be used to update the LAPIS filter and which to update the value?

components/src/web-components/visualization/introduction.mdx
anna-parker: The input LAPIS filter should only contain a LAPIS query filter (i.e. the input query should not include request parameters such as `fields`, `limit`, `orderBy`).
The components perform minor modifications to the LAPIS filter before sending it to LAPIS, such as adding stratification by adding which `fields` the query should return.

components/src/web-components/input/introduction.mdx
fengelniederhammer: `CustomEvent`s always have the `detail` property. I added a sentence to check the docs of the individual components. Do you think that helps?

anna-parker: Hmm... so I still feel like there is a contradiction in the text:

1. Every component fires an event that can be used to update the LAPIS filters
2. every component fires an event whose `detail` can be used as the `value` of the component.

But then you say "Some components only fire a single event when the event can be used to update the LAPIS filter **and** the value of the component." which implies the other components events can't be used to (1) update LAPIS filters and (2) update the value - so this seems like a contradiction to me. 

It sounds like 1 and 2 do not always hold

components/src/web-components/visualization/introduction.mdx
```diff
@@ -0,0 +1,51 @@
...
+### Scales
+
+On most plots, users can select the y-axis scaling through a dropdown.
+They can choose between linear, logarithmic and logistic scaling.
+By default, it is set to a linear scale.
+
+In general, for each scale the displayed height of a value is calculated by applying the corresponding scale function.
+
+- Linear: `value`
+- Logarithmic: `ln(value)`
```
anna-parker, line 41: maybe mention how log of zero is handled

anna-parker (approved):

anna-parker:
Ah and if you could fix the format issues that would be great :-)

PR #921, opened 2025-07-05
fix(components): accept lineage definitions without "aliases" key

### LINKED ISSUE #920: Lineage filter doesn't work if no aliases are defined
If a lineage system doesn't have aliases (e.g., dengue), LAPIS won't return `aliases` in the `/sample/lineageDefinition/{column}` endpoint. This is causing the `lineage-filter` to error:

```
{
    "code": "invalid_type",
    "expected": "array",
    "received": "undefined",
    "path": [
      "2III",
      "aliases"
    ],
    "message": "Required"
}
```

chaoran-chen:
resolves #920

This PR make the components to accept lineage definition files from LAPIS where `aliases` are not defined.

### PR Checklist

- ~[ ] All necessary documentation has been adapted.~

### DIFF
--- a/components/src/lapisApi/LineageDefinition.ts
+++ b/components/src/lapisApi/LineageDefinition.ts
@@ -2,7 +2,7 @@ import z from 'zod';
 
 const lineage = z.object({
     parents: z.array(z.string()).optional(),
-    aliases: z.array(z.string()),
+    aliases: z.array(z.string()).optional(),
 });
 
 export const lineageDefinitionResponseSchema = z.record(lineage);
--- a/components/src/preact/lineageFilter/fetchLineageAutocompleteList.spec.ts
+++ b/components/src/preact/lineageFilter/fetchLineageAutocompleteList.spec.ts
@@ -47,6 +47,44 @@ describe('fetchLineageAutocompleteList', () => {
         ]);
     });
 
+    test('should work without aliases', async () => {
+        lapisRequestMocks.aggregated(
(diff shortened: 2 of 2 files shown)


components/src/lapisApi/LineageDefinition.ts
```diff
@@ -2,8 +2,14 @@ import z from 'zod';
...
     parents: z.array(z.string()).optional(),
-    aliases: z.array(z.string()),
+    aliases: z.array(z.string()).default([]),
 });
 
-export const lineageDefinitionResponseSchema = z.record(lineage);
+// LAPIS' response does not include `aliases` as a key if no aliases are defined
+const lineageResponse = lineage.extend({
+    aliases: lineage.shape.aliases.optional(),
+});
```
fengelniederhammer, line 11: Why do we need `lineage` and `lineageResponse`? Isn't one of them enough?

I would prefer to use `lineageResponse`. I don't like the default in the type definition because it could be surprising that the runtime check of a type changes the value (namely turning `aliases: undefined' into `aliases: []`). I'd rather be specific and handle the undefined case in our code somewhere.

(Also in the current state the test code uses `LineageDefinitionResponse`, but the runtime code uses `lineageDefinitionSchema` which is extra weird)

fengelniederhammer (changes requested):

fengelniederhammer:
If you add `resolves #920` in the commit message during merge, then the issue should also show up in the changelog.

PR #927, opened 2025-07-08
feat(components): gs-mutations-over-time: add checkboxes to quickly filter by annotated mutations

### LINKED ISSUE #911: Easier access to annotations
When presenting the dashboards it would be really nice to have a button to click on to filter to all the mutations that are annotated - this could be next to the filter mutations box

Image

fengelniederhammer:
resolves #911

### Summary

This adds a checkbox for every annotation to the "filter mutations" dropdown. The checkboxes are concatenated with logical "or" so that checking several checkboxes will show entries of all checked annotations. The text filter is concatenated with the checkboxes with "and", so that it will only show entries that are annotated and match the text filter.

### Screenshot

image

### PR Checklist


components/src/preact/mutationsOverTime/getFilteredMutationsOverTimeData.ts
```diff
@@ -72,25 +77,54 @@ export function getFilteredMutationOverTimeData({
 export function mutationOrAnnotationDoNotMatchFilter(
     mutation: Mutation,
     sequenceType: SequenceType,
-    filterValue: string,
+    mutationFilter: MutationFilter,
     annotationProvider: ReturnType<typeof useMutationAnnotationsProvider>,
 ) {
-    if (filterValue === '') {
-        return false;
+    return !(
```
anna-parker, line 83: isn't this wrong? we want to keep a mutation (not filter out) if `mutationOrAnnotationMatchesTextFilter` OR `mutationMatchesAnnotationFilter`
so the reverse of this A OR B is NOT(A OR B) which is (NOT A) && (NOT B) this is not equivalent to NOT (A && B)

anna-parker: Ah ok I just was confused as the `mutationMatchesAnnotationFilter` does OR filtering... I find this behavior a bit unexpected I would either expect all OR or all AND. But maybe this is logical for others? Could we maybe ask in a channel? I might be wrong here

anna-parker:
Anika just responded that this implementation looks good - thanks for this @fengelniederhammer!

PR #942, opened 2025-08-06
feat(components): use new mutationsOverTime LAPIS endpoint in the mutations-over-time component

fhennig:
### Summary
This PR implements a new method to fetch the data for the _MutationsOverTime_ component. A new LAPIS endpoint was implemented which directly provides the data that the plot needs (see https://github.com/GenSpectrum/LAPIS/pull/1229).
The new endpoint can be used in the component with a boolean switch called `useNewEndpoint`, it defaults to `false`.

With the new endpoint, the plot behaves a little bit different:
- the new endpoint does not generate `null` and `belowThreshold`cells anymore.
- when selecting and end date like 2022-08-12 and selecting granularity monthly, the old calculating method presented counts for the whole month of August, whereas the new endpoint only shows data up to the selected `dateTo`. This is actually more in line with how other plots work.

#### Testing
I've added a new dedicated test file for the new endpoint.

I've manually checked that the storybook demos still work when setting the new parameter to true.

### Screenshot
n/a

### PR Checklist

### DIFF
--- a/components/src/lapisApi/lapisApi.ts
+++ b/components/src/lapisApi/lapisApi.ts
@@ -9,6 +9,8 @@ import {
     mutationsResponse,
     problemDetail,
     type ProblemDetail,
+    type MutationsOverTimeRequest,
+    mutationsOverTimeResponse,
 } from './lapisTypes';
 import { type SequenceType } from '../types';
 import { lineageDefinitionResponseSchema } from './LineageDefinition';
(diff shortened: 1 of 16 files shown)


components/src/lapisApi/lapisApi.ts
chaoran-chen, line 1: This should be classified as `feat` it will change the handling of low proportions and the performance (in a non-trivial way; sometimes better, sometimes worse?).

Also, this should be scoped with `components`, see [.commitlintrc](https://github.com/GenSpectrum/dashboard-components/blob/main/.commitlintrc.js#L9-L19)

fhennig: Ah thanks, I knew I'd have to edit the title again, wasn't sure about the allowed scopes etc

PR #948, opened 2025-08-12
fix(components): set bg color as white in fullscreen

### LINKED ISSUE #930: Black blackground in fullscreen
If I switch on the fullscreen mode, the components get a black background, making them unreadable. For example, for mutations over time:

Image

fhennig:
resolves #930 

### Work in progress - thoughts

I have a simple fix: I'm just setting the background color of our tabs container; we use it everywhere, and that works:

image

But further research showed that there's also `#container:backdrop` which one can set (https://stackoverflow.com/questions/16163089/background-element-goes-black-when-entering-fullscreen-with-html5) - and I think we could possibly add this as a custom class and then attach it dynamically to the fullscreened element. Then, we could fullscreen arbitrary components.

Would that be "better"? Arguably it's simpler to just set the bg color on the root element.

I feel like the user of the component library would be in charge of setting the background color on the fullscreen panel, it's not really our job of the component. So maybe it's the proper way to go to specify a bg color for our components?

### Summary
This PR sets a fixed background color for the `Tabs` element, so when it is fullscreened, it still has a white background and is readable.

### Testing
I didn't add specific tests yet - do we want tests that check that components are readable in fullscreen?

### Screenshot
image

### PR Checklist

### DIFF
--- a/components/src/preact/components/tabs.tsx
+++ b/components/src/preact/components/tabs.tsx
@@ -40,7 +40,7 @@ const Tabs = forwardRef<HTMLDivElement, ComponentTabsProps>(({ tabs, toolbar }, 
     const toolbarElement = typeof toolbar === 'function' ? toolbar(activeTab) : toolbar;
 
     return (
-        <div ref={ref} className='h-full w-full flex flex-col'>
+        <div ref={ref} className='h-full w-full flex flex-col bg-white'>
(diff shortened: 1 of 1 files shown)


chaoran-chen (approved):
I think this solution is great and can be merged! It is entirely fine for the components to define a background as they will only work well with a white one and there is no need for tests in this case.

PR #956, opened 2025-08-19
docs(components): document test runner install info for MacOS

### LINKED ISSUE #949: package-lock.json contains Linux-specific storybook test-runner deps; doesn't work on Mac
Running `npm run test:storybook` on MacOS (after `npm ci`) gives me this error:

```
Error: Failed to load native binding
    at Object.<anonymous> (/Users/felix/loc/dashboard-components/components/node_modules/@swc/core/binding.js:329:11)
    at Module._compile (node:internal/modules/cjs/loader:1739:14)
    at Object..js (node:internal/modules/cjs/loader:1904:10)
    at Module.load (node:internal/modules/cjs/loader:1473:32)
    at Function._load (node:internal/modules/cjs/loader:1285:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:234:24)
    at Module.require (node:internal/modules/cjs/loader:1495:12)
    at require (node:internal/modules/helpers:135:16)
    at Object.<anonymous> (/Users/felix/loc/dashboard-components/components/node_modules/@swc/core/index.js:49:17)
```

The `package-lock.json` has `"node_modules/@swc/core-linux-x64-gnu"` and `"node_modules/@swc/core-linux-x64-musl"`, but for MacOS we would need `"node_modules/@swc/core-darwin-arm64"`.

To fix this, it works to reinstall the test-runner:

```sh
npm uninstall @storybook/test-runner
npm install @storybook/test-runner --force
```

Can we somehow add multiple optional arch-specific deps to the package-lock? Maybe for now, it'd be good to add info to the README that you need to reinstall the test-runner to run the tests.

----

I also had to change this line: https://github.com/GenSpectrum/dashboard-components/blob/main/components/.storybook/test-runner.ts#L2 to import with `.ts`, because otherwise I got an error that the file wasn't found. But maybe that's just a local issue I'm having.

fhennig:
related to #949

### Summary
Document that you need to reinstall the test-runner for MacOS to make it work.

### Screenshot
image

### PR Checklist

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -77,6 +77,15 @@ We primarily provide two kinds of components:
 npm ci
 ```
 
+> [!NOTE]  
+> **For Mac users**: By default, the `package-lock.json` installs linux dependencies for the
+> storybook test runner. To use it, you need to reinstall it:
+>
+> ```
+> npm uninstall @storybook/test-runner
+> npm install @storybook/test-runner --force
+> ```
+
 ### Custom Elements Manifest
 
 This package also ships a [Custom Elements Manifest](https://custom-elements-manifest.open-wc.org/),


components/README.md
```diff
@@ -77,6 +77,15 @@ We primarily provide two kinds of components:
 npm ci
```
fengelniederhammer, line 77: I just saw that this says you should use `npm ci`. Just to be sure: What happens when you use `npm i` instead?

fengelniederhammer (approved):

components/README.md
fhennig: If I remember correctly, the test runner works, but there was an issue with running the storybook. 

Also, isn't it better to actually install what's in the lockfile?

PR #973, opened 2025-09-03
fix(components): display correct mutation codes when using incomplete mutation codes in the displayMutations

fhennig:
### Summary

When giving, for example, just '405T' as a mutation, the response will include 'A405T'. We want to use that code, instead of the incomplete one.

### Screenshot
n/A

### PR Checklist

### DIFF
--- a/components/src/query/queryMutationsOverTime.ts
+++ b/components/src/query/queryMutationsOverTime.ts
@@ -267,12 +267,33 @@ async function queryMutationsOverTimeDataDirectEndpoint(
         signal,
     );
 
+    const responseMutations = apiResult.data.mutations.map(parseMutationCode);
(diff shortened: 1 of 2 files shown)


components/src/query/queryMutationsOverTime.ts
```diff
@@ -267,12 +267,33 @@ async function queryMutationsOverTimeDataDirectEndpoint(
...
+        if (mutation.type === 'deletion') {
+            return {
+                type: 'deletion',
+                mutation,
+                ...numbers,
+            };
+        } else {
+            return {
+                type: 'substitution',
+                mutation: mutation,
```
chaoran-chen, line 285:
```suggestion
                mutation,
```

chaoran-chen (approved):

PR #977, opened 2025-09-05
feat(components): allow disabling counts

### LINKED ISSUE #970: Allow disabling result counts in filters
The filters show these helpful number counts:

Image

I'm pulling in lineage information from a different lapis instance, so the numbers are not relevant to me. I'd like to be able to disable them. I need this only on the `gs-lineage-filter` but I think it'd be good to add this option everywhere for consistency.

fhennig:
resolves #970 

### Summary
Adds a `showCounts` prop to the filter components (Lineage, Location, Text) to allow enabling/disabling the grey counts behind the options. It defaults to true, so behavior is unchanged.

I've added storybook tests.

### Screenshot

image

### PR Checklist

### DIFF
--- a/components/src/preact/lineageFilter/lineage-filter.stories.tsx
+++ b/components/src/preact/lineageFilter/lineage-filter.stories.tsx
@@ -74,6 +74,11 @@ const meta: Meta = {
                 type: 'object',
             },
         },
+        hideCounts: {
+            control: {
+                type: 'boolean',
+            },
+        },
     },
 
     args: {
@@ -84,6 +89,7 @@ const meta: Meta = {
         placeholderText: 'Enter a lineage',
         value: 'A.1',
         width: '100%',
+        hideCounts: false,
     },
 };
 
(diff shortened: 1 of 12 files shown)


fengelniederhammer (changes requested):
Unfortunately boolean attributes are somewhat tricky. If they are present, then they are `true`. They can only be set to `false` by omitting them.
https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML

```
                <gs-lineage-filter
                    showCounts="false"
                ></gs-lineage-filter>
```
will still show the counts.

This however works:
```
                html`<gs-lineage-filter
                    .showCounts=${args.showCounts}
                ></gs-lineage-filter>`
```
(that's why you didn't see that in the tests). It's the lit HTML template method that makes it work. By using the `.` before the attribute, it doesn't actually set the HTML attribute, it rather sets the property on the underlying class directly (Surprise! I don't like this method due to that reason.).
https://lit.dev/docs/v1/lit-html/writing-templates/#bind-to-properties
That's also the reason why you don't see the attributes in the rendered HTML in the stories (since they are not set in the HTML):
image

I think we need to invert the attributes to `hideCounts` with a default of `false`. (Edit: Since we need to be able to use the compnoents in plain HTML, not just with this weird Lit HTML template function)

fhennig:
Ah, I didn't notice! Good catch. Okay, yes, inverting to hideCounts sounds good to me I think, I'll go ahead with that.

fengelniederhammer (approved):
LGTM

PR #980, opened 2025-09-09
feat(components): mutation filter: Make type of mutations configurable

### LINKED ISSUE #963: mutations filter: make it configurable which kinds of mutations can be entered
I'd like to use this component, but only support entering nucleotide mutations. no insertions, no amino acid mutations.

Maybe there could be a list of `enabledInputTypes` which could contain `nucleotideMutation`, `nucleotideInsertion`, `aminoAcidMutation`, `aminoAcidInsertion`. By default, its all of them (same as now) but I could also select just the ones I want.

fhennig:
resolves #963 

### Summary

Adds a new property `enabledMutationTypes` which is a list of strings, which can have these:

 - `nucleotideMutations`
 - `nucleotideInsertions`
 - `aminoAcidMutations`
 - `aminoAcidInsertions`

The help modal has not been adjusted, I believe it is not necessary to adjust the text.

#### Testing

I've added a few spot tests to the storybook tests

### Screenshot

https://github.com/user-attachments/assets/c76b30c7-9b6f-4b79-b154-2ac45c7bf8cf

### PR Checklist

### DIFF
--- a/components/src/preact/mutationFilter/mutation-filter.tsx
+++ b/components/src/preact/mutationFilter/mutation-filter.tsx
@@ -1,6 +1,6 @@
 import { useCombobox, useMultipleSelection } from 'downshift/preact';
 import { type FunctionComponent } from 'preact';
-import { useContext, useMemo, useRef, useState } from 'preact/hooks';
+import { useContext, useEffect, useMemo, useRef, useState } from 'preact/hooks';
 import z from 'zod';
 
 import { getExampleMutation } from './ExampleMutation';
@@ -15,8 +15,18 @@ import { ErrorBoundary } from '../components/error-boundary';
 import { UserFacingError } from '../components/error-display';
 import { singleGraphColorRGBByName } from '../shared/charts/colors';
 
+const mutationTypeSchema = z.enum([
+    'nucleotideMutations',
+    'aminoAcidMutations',
+    'nucleotideInsertions',
+    'aminoAcidInsertions',
+]);
+
+export type MutationType = z.infer<typeof mutationTypeSchema>;
+
 const mutationFilterInnerPropsSchema = z.object({
     initialValue: z.union([mutationsFilterSchema.optional(), z.array(z.string()), z.undefined()]),
+    enabledMutationTypes: z.array(mutationTypeSchema).optional(),
 });
 
 const mutationFilterPropsSchema = mutationFilterInnerPropsSchema.extend({
@@ -53,7 +63,7 @@ export type MutationFilterItem =
     | SelectedAminoAcidInsertion;
 
 export const MutationFilter: FunctionComponent<MutationFilterProps> = (props) => {
-    const { width, initialValue } = props;
+    const { width, initialValue, enabledMutationTypes } = props;
     return (
         <ErrorBoundary
             size={{ height: '40px', width }}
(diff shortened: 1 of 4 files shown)


components/src/web-components/input/gs-mutation-filter.tsx
```diff
@@ -52,6 +56,8 @@ import { PreactLitAdapter } from '../PreactLitAdapter';
  * Fired when:
  * - The user has submitted a valid mutation or insertion
  * - The user has removed a mutation or insertion
+ *
+ * TODO - update this docs description to mention the new field
```
fengelniederhammer, line 60: Reminder :)

components/src/web-components/input/gs-mutation-filter.tsx
```diff
@@ -81,10 +87,28 @@ export class MutationFilterComponent extends PreactLitAdapter {
...
+     *
+     * - `nucleotideMutations`
+     * - `nucleotideInsertions`
+     * - `aminoAcidMutations`
+     * - `aminoAcidInsertions`
+     *
+     * By default or if none are given, all types are accepted.
+     */
+    @property({ type: Object })
+    enabledMutationTypes: MutationType[] | undefined = undefined;
```
fengelniederhammer, line 102: Usually, we spelled out the types to primitive types so that they are clear to someone who just looks at the storybook docs since there it just shows the type name, but not what the type is:
image

But I think since you mentioned the available values, it should be clear enough in this case.

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -115,16 +128,24 @@ function MutationFilterInner({ initialValue }: MutationFilterInnerProps) {
...
-            const invalidInput = values
-                .filter((value) => value.parsedValue === null)
-                .map((value) => value.value.trim())
-                .join(',');
+
+            const validEntries: MutationFilterItem[] = [];
+            const rejected: string[] = [];
+
+            for (const v of values) {
+                if (!v.parsedValue) {
```
fengelniederhammer, line 136:
```suggestion
                if (v.parsedValue === null) {
```
Let's please be explicit to avoid subtle edge cases with things that are falsy.

components/src/preact/mutationFilter/mutation-filter.tsx
```diff
@@ -133,7 +154,8 @@ function MutationFilterInner({ initialValue }: MutationFilterInnerProps) {
                 const alreadyExists = selectedItems.find(
                     (selectedItem) => selectedItem.value.code === candidate?.value.code,
                 );
-                if (!alreadyExists) {
+                const allowedType = candidate && enabledMutationTypes.includes(candidate.type);
```
fengelniederhammer, line 157:
```suggestion
                const allowedType = candidate !== null && enabledMutationTypes.includes(candidate.type);
```

fengelniederhammer (commented):
I found a weird edge case:
image

* Allow AA mutations
* enter an AA mutation
* change the attributes to not allow AA mutations
-> The AA mutation is still there but it doesn't allow any new ones.

Is this important enough to consider? I believe setting a key somewhere that rerenders everything when the `enabledMutationTypes` change would be a proper fix?

fengelniederhammer (approved):
LGTM

PR #995, opened 2025-09-23
feat(components): add mutation link templates

### LINKED ISSUE #985: mutations over time: custom render function for mutation names
I'd like to be able to make the mutations clickable, and link them to cov-spectrum. In the future, we'd likely also like to colorize the mutations.
For that I was thinking maybe we can just make it possible for the user to pass in a custom rendering function, that gets the mutation and context info like the annotations.

Image

fhennig:
resolves #985 

### Summary

Adds a new attribute to `gs-app`, `mutationLinkTemplate`:

```html
<gs-app
    lapis="https://your.lapis.url"
    mutationLinkTemplate="{
        nucleotideMutation: 'https://cov-spectrum.org/explore/Switzerland/AllSamples/Past6M/variants?nucMutations={{mutation}}',
        aminoAcidMutation: 'https://cov-spectrum.org/explore/Switzerland/AllSamples/Past6M/variants?aaMutations={{mutation}}'
    }"
>
    {/* children... */}
</gs-app>
```

where you can put a link with the placeholder `{{mutation}}` which will then be replaced by the mutation code.

The `AnnotatedMutation` component has been extended to also take the links into account.

Annotations are now only clickable by clicking the annotation itself, not the whole mutation.

#### Testing

added new tests to the `storybook-preact`

### Screenshot

image

### PR Checklist

### DIFF
--- a/components/src/web-components/gs-app.ts
+++ b/components/src/web-components/gs-app.ts
@@ -7,6 +7,7 @@ import z from 'zod';
 
 import { lapisContext } from './lapis-context';
 import { mutationAnnotationsContext } from './mutation-annotations-context';
+import { mutationLinkTemplateContext } from './mutation-link-template-context';
 import { referenceGenomeContext } from './reference-genome-context';
 import { type ReferenceGenome } from '../lapisApi/ReferenceGenome';
 import { fetchReferenceGenome } from '../lapisApi/lapisApi';
@@ -58,6 +59,22 @@ export class AppComponent extends LitElement {
         aminoAcidMutations?: string[];
         aminoAcidPositions?: string[];
     }[] = [];
+
+    /**
+     * Supply a link template for nucleotide and amino acid mutations.
+     * The template should include '{{mutation}}' where the mutation code will be inserted, for example:
+     *
+     *     https://my-site.org/query?nucleotideMutation={{mutation}}
+     */
+    @provide({ context: mutationLinkTemplateContext })
+    @property({ type: Object })
+    mutationLinkTemplate: {
+        nucleotideMutation?: string;
+        aminoAcidMutation?: string;
+    } = {
+        nucleotideMutation: undefined,
+        aminoAcidMutation: undefined,
+    };
 
     /**
      * @internal
--- a/components/src/web-components/mutation-link-template-context.ts
+++ b/components/src/web-components/mutation-link-template-context.ts
@@ -1,0 +1,13 @@
+import { createContext } from '@lit/context';
+import z from 'zod';
+
+export const mutationLinkTemplateSchema = z.object({
+    nucleotideMutation: z.string().optional(),
+    aminoAcidMutation: z.string().optional(),
+});
+
+export type MutationLinkTemplate = z.infer<typeof mutationLinkTemplateSchema>;
+
+export const mutationLinkTemplateContext = createContext<MutationLinkTemplate>(
+    Symbol('mutation-link-template-context'),
+);
--- a/components/src/web-components/mutationLinks.mdx
+++ b/components/src/web-components/mutationLinks.mdx
@@ -1,0 +1,27 @@
+import { Meta } from '@storybook/blocks';
+
+<Meta title='Concepts/External Mutation Links' />
+
+# External Mutation Links
+
+When inspecting mutations, you might want to be able to quickly look up a mutation somewhere else.
+For this, it's possible to supply a _link template_ which will then lead to mutation labels being clickable using
+that link template.
+
+The mutation link template can be (optionally) supplied to `gs-app` as a JSON object:
+
+```html
+<gs-app
+    lapis="https://your.lapis.url"
+    mutationLinkTemplate='{
+        "nucleotideMutation": "https://cov-spectrum.org/explore/Switzerland/AllSamples/Past6M/variants?nucMutations={{mutation}}",
+        "aminoAcidMutation": "https://cov-spectrum.org/explore/Switzerland/AllSamples/Past6M/variants?aaMutations={{mutation}}",
+    }'
+>
+    {/* children... */}
+</gs-app>
+```
+
+For example (like in the code above) you can directly link to cov-spectrum.
+
+Your link needs to have the placeholder `{{mutation}}` in it, which will then be replaced with the mutation code.
--- a/components/src/preact/MutationLinkTemplateContext.tsx
+++ b/components/src/preact/MutationLinkTemplateContext.tsx
@@ -1,0 +1,56 @@
+import { createContext, type Provider } from 'preact';
+import { useContext, useMemo } from 'preact/hooks';
+
+import type { SequenceType } from '../types';
+import type { Deletion, Substitution } from '../utils/mutations';
+import { mutationLinkTemplateSchema } from '../web-components/mutation-link-template-context';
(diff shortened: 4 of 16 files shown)


components/src/web-components/gs-app.ts
```diff
@@ -59,6 +60,22 @@ export class AppComponent extends LitElement {
...
 
+    /**
+     * Supply a link template for nucleotide and amino acid mutations.
+     * The template should include '{{mutation}}' where the mutation code will be inserted, for example:
+     *
+     *     https://my-site.org/query?nucleotideMutation={{mutation}}
+     */
+    @provide({ context: mutationLinkTemplateContext })
+    @property({ type: Object })
+    mutationLinkTemplate: {
```
fengelniederhammer, line 71: Let's add a section to the dummy component (`AppDisplay`) in the story that visualizes the value here? Then we could also add a test that checks that the values correctly finds its way into the component.

components/src/web-components/gs-app.ts
```diff
@@ -59,6 +60,22 @@ export class AppComponent extends LitElement {
...
+        nucleotideMutation?: string;
+        aminoAcidMutation?: string;
+    } = {
+        nucleotideMutation: undefined,
+        aminoAcidMutation: undefined,
+    };
```
fengelniederhammer, line 77: Let's add a test to `gs-app.spec-d.ts` that checks that the type matches the type that the `MutationLinkTemplateContext` expects?

components/src/web-components/mutation-link-template-context.ts
```diff
@@ -0,0 +1,13 @@
+import { createContext } from '@lit/context';
+import z from 'zod';
+
+const mutationLinkTemplateSchema = z.object({
```
fengelniederhammer, line 4: We should use this schema somewhere to check the context value.

components/src/web-components/mutationLinks.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+The mutation link template can be (optionally) supplied to `gs-app` as a JSON object:
+
+```html
+<gs-app
+    lapis="https://your.lapis.url"
+    mutationLinkTemplate="{
+        nucleotideMutation: 'https://cov-spectrum.org/explore/Switzerland/AllSamples/Past6M/variants?nucMutations={{mutation}}',
+        aminoAcidMutation: 'https://cov-spectrum.org/explore/Switzerland/AllSamples/Past6M/variants?aaMutations={{mutation}}'
+    }"
```
fengelniederhammer, line 19: This is not a valid JSON ;) It's missing the `"` around the properties. And the values should also use `"` instead of  `'`.

I think you can use `'` for the attribute instead so that you can use the `"` inside without escaping.

components/src/preact/MutationLinkTemplateContext.ts
```diff
@@ -0,0 +1,40 @@
...
+
+            case 'amino acid': {
+                if (linkTemplate.aminoAcidMutation !== null) {
+                    link = linkTemplate.aminoAcidMutation?.replace('{{mutation}}', encodeURIComponent(mutation.code));
+                }
+                break;
+            }
+        }
+
+        return link;
```
fengelniederhammer, line 38: Do we need `link`?

```suggestion
        switch (sequenceType) {
            case 'nucleotide': {
                if (linkTemplate.nucleotideMutation !== null) {
                    return linkTemplate.nucleotideMutation?.replace('{{mutation}}', encodeURIComponent(mutation.code));
                }
                return undefined;
            }

            case 'amino acid': {
                if (linkTemplate.aminoAcidMutation !== null) {
                    return linkTemplate.aminoAcidMutation?.replace('{{mutation}}', encodeURIComponent(mutation.code));
                }
                return undefined;
            }
        }
```

components/src/web-components/mutation-link-template-context.ts
```diff
@@ -0,0 +1,13 @@
...
+    nucleotideMutation: z.string().optional(),
+    aminoAcidMutation: z.string().optional(),
+});
+
+export type MutationLinkTemplate = z.infer<typeof mutationLinkTemplateSchema>;
+
+export const mutationLinkTemplateContext = createContext<MutationLinkTemplate>(
```
fengelniederhammer, line 11: I think we could also use that context everywhere where we use the `mutationAnnotationsContext`. They essentially have the same target, don't they?

components/src/web-components/mutation-link-template-context.ts
fhennig: I had a look at how the schema is used for the annotations, and found it to be used here: https://github.com/GenSpectrum/dashboard-components/blob/link-templates/components/src/preact/MutationAnnotationsContext.tsx#L39

I found it actually to be not the right place, because seemingly, the incoming value should already have the correct type. I looked for ways to parse already earlier, and made this commit: https://github.com/GenSpectrum/dashboard-components/pull/995/commits/8cfe15e3c9fcb82b411643011da055ee05e58d96

It uses the 'converter' on a lit property to parse directly the string into the correct type, and return an empty object, and raise a console error if the parsing fails. I think this is the way to go, maybe we should also refactor the annotations property to do it like this? because if we don't already validate, we have an object in the system that doesn't actually have the correct type at runtime!

components/src/web-components/mutation-link-template-context.ts
fhennig: https://github.com/GenSpectrum/dashboard-components/pull/995/commits/d5707d9dfef8fb6d396458f1b7b3a30e779c1a81

components/src/web-components/gs-app.ts
```diff
@@ -59,6 +60,25 @@ export class AppComponent extends LitElement {
...
+    mutationLinkTemplate: MutationLinkTemplate = {
+        nucleotideMutation: undefined,
+        aminoAcidMutation: undefined,
+    };
+
+    /**
+     * @internal
+     */
+    @property({ type: String })
+    validationError = '';
```
fengelniederhammer, line 80: I suspect that we don't need this anymore?

components/src/web-components/gs-app.ts
```diff
@@ -59,6 +60,25 @@ export class AppComponent extends LitElement {
...
 
+    /**
+     * Supply a link template for nucleotide and amino acid mutations.
+     * The template should include '{{mutation}}' where the mutation code will be inserted, for example:
+     *
+     *     https://my-site.org/query?nucleotideMutation={{mutation}}
+     */
+    @provide({ context: mutationLinkTemplateContext })
+    @property({ type: Object })
+    mutationLinkTemplate: MutationLinkTemplate = {
```
fengelniederhammer, line 71: I think we should spell that type out so that it shows up usefully in storybook. Currently, someone who doesn't look at our code will not understand the type: 
Image


```suggestion
    mutationLinkTemplate: { nucleotideMutation: string | undefined, aminoAcidMutation: string l undefined } = {
```

fengelniederhammer (commented):
Thanks, we're almost there!

components/src/web-components/gs-app.ts
fhennig: https://github.com/GenSpectrum/dashboard-components/pull/995/commits/2c24df2c9c45a8cc7333c37c408f77fdca996f32

fengelniederhammer (approved):

PR #996, opened 2025-09-24
chore(components): export utility types

### LINKED ISSUE #983: mutations over time: export `InitialMeanPropInterval` as a type
export this: https://github.com/GenSpectrum/dashboard-components/blob/4707e3807d5b05a4c6ee20751caa6ee72a94d98b/components/src/preact/mutationsOverTime/mutations-over-time.tsx#L58 as a type to use by users of the component

fhennig:
resolves #983 , also relevant for https://github.com/GenSpectrum/dashboards/pull/833

### Summary

Export `MeanProportionInterval` and `MutationType`.

Add a little README snippet about exporting types.

### Screenshot
n/a

### PR Checklist

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -178,3 +178,7 @@ workers that return mocked data. The mock workers are enabled in the package.jso
 Node.js [subpath imports](https://nodejs.org/api/packages.html#subpath-imports), following the guide
 from [storybook](https://storybook.js.org/docs/writing-stories/mocking-data-and-modules/mocking-modules). This ensures
 that when importing the worker in the component, the mock worker is used inside Storybook instead of the real worker.
+
+### Exporting types into the final build
+
+This project uses `vite build`. The release config is `vite.release.config.ts`. In there, the lib export entrypoints are defined. If you want to export a utility type for your component, re-export it inside of `utilEntrypoint.ts`.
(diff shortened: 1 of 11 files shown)


components/README.md
```diff
@@ -178,3 +178,7 @@ workers that return mocked data. The mock workers are enabled in the package.jso
 Node.js [subpath imports](https://nodejs.org/api/packages.html#subpath-imports), following the guide
 from [storybook](https://storybook.js.org/docs/writing-stories/mocking-data-and-modules/mocking-modules). This ensures
 that when importing the worker in the component, the mock worker is used inside Storybook instead of the real worker.
+
+### Exporting types into the final build
+
+This project uses `vite build`. The release config is `vite.release.config.ts`. In there, the lib export entrypoints are defined. If you want to export a utility type for your component, re-export it inside of `utilEntrypoint.ts`.
```
fengelniederhammer, line 184: I think we could add something more general about how the build and the publishing works. But this is already an improvement, so I won't block 👍

fengelniederhammer (approved):

components/README.md
fhennig: Yes I agree, just didn't know exactly what to write. I think you're maybe better suited to write a blurp about that, that would be great!

fhennig: I made an issue for it: https://github.com/GenSpectrum/dashboard-components/issues/997 - maybe you can pick it up if you have some time

PR #1019, opened 2025-10-27
feat(components): custom column support for mutations-over-time

fhennig:
possibly a solution for https://github.com/GenSpectrum/dashboards/issues/886

### Summary

Allows the user of the component to provide a map of mutation codes to values which will then be displayed in a dedicated new column.

One use-case is showing the Jaccard index for a particular mutation based on a particular variant.

### Screenshot

image

### PR Checklist


fengelniederhammer:
> Update: I just had a Chat with @chaoran-chen about this and we thought that this is probably a good feature to go forward with.
> 
> We discussed the issue of there not being a cell value possibly for every mutation if you do not give an explicit list of `includeMutations`, but that would just be similar to how the annotations work, not every row will have a value, which isn't a problem.
> 
> @fengelniederhammer we wanted to get your input again if have other thoughts about this.

Sounds good to me.

components/src/web-components/visualization/gs-mutations-over-time.tsx
```diff
@@ -136,6 +136,13 @@ export class MutationsOverTimeComponent extends PreactLitAdapterWithGridJsStyles
     @property({ type: Array })
     pageSizes: number[] | number = [10, 20, 30, 40, 50];
 
+    /**
+     * Custom columns to add to the grid.
+     * Each column has a header and a map of mutation codes to values.
+     */
+    @property({ type: Array })
+    customColumns?: { header: string; values: Record<string, string | number> }[] = undefined;
```
fengelniederhammer: Let's add the column to the type test.

components/src/preact/mutationsOverTime/mutations-over-time.stories.tsx
```diff
@@ -196,6 +197,36 @@ export const UsesMutationFilter: StoryObj<MutationsOverTimeProps> = {
...
+                values: {
+                    A19722G: 0.75,
+                    G21641T: "Foobar",
+                },
+            },
+        ],
+    },
+    play: async ({ canvas }) => {
+        await waitFor(() => expect(canvas.getByText('Jaccard Index')).toBeVisible(), {
+            timeout: 10000,
```
fengelniederhammer: Why do we need the extra long timeout here?

components/src/web-components/visualization/gs-mutations-over-time.tsx
fengelniederhammer: Should we also allow booleans?

fhennig: I mean, in theory we can just have whatever in the column, right? We don't really interact with the values in the column. 

I'd say we just leave it the way it is for now, it's easy to add more stuff later.

fengelniederhammer (approved):

PR #1020, opened 2025-10-28
fix(components): mutations-over-time tooltip overflow without scroll

### LINKED ISSUE #981: mutations over time: fix scroll overflow
The mutations-over-time component has weird overflow issues.

#### Genspectrum

https://github.com/user-attachments/assets/2c11591d-10ce-44e2-aeb6-604cbd7f0e95

#### Storybook

https://github.com/user-attachments/assets/38eb0114-f937-4944-b5bf-1517ab553548

### LINKED ISSUE #984: mutations over time: tooltip not shown completely for few rows
On [this staging page](https://staging.genspectrum.org/swiss-wastewater/covid?granularity=week&analysisMode=manual&sequenceType=amino+acid&mutations=S%3AK356%7CS%3A501%7CS%3A69-&) where a user might choose only very few mutations, the tooltips are vastly not visible.

It might not be easy to fix: We have a related issue in #391 and had in the past #634. Nonetheless, it would be good to discuss it again and explore options.

Image

fhennig:
resolves #984, resolves #981 

### Summary

Adds a new `PortalTooltip` that uses a portal internally to position the tooltip. In the mutations-over-time component and the grid, we're passing through an outer div ref where we can render the tooltip to.

This fixes the issues, because the tooltip is no longer inside the tab, so it doesn't cause the overflow and scroll bars anymore. The portal is also still inside the shadow DOM, so it doesn't break out of the shadow DOM.

### Screenshot

https://github.com/user-attachments/assets/afd2c50b-c90b-478f-be44-925955510631

### PR Checklist

### DIFF
--- a/components/src/preact/mutationsOverTime/mutations-over-time.tsx
+++ b/components/src/preact/mutationsOverTime/mutations-over-time.tsx
@@ -1,5 +1,5 @@
 import { type FunctionComponent } from 'preact';
-import { type Dispatch, type StateUpdater, useMemo, useState, useEffect, useRef } from 'preact/hooks';
+import { type Dispatch, type StateUpdater, useMemo, useState, useEffect, useLayoutEffect, useRef } from 'preact/hooks';
 import z from 'zod';
 
 // @ts-expect-error -- uses subpath imports and vite worker import
@@ -143,6 +143,11 @@ const MutationsOverTimeTabs: FunctionComponent<MutationOverTimeTabsProps> = ({
 }) => {
     const tabsRef = useDispatchFinishedLoadingEvent();
     const tooltipPortalTargetRef = useRef<HTMLDivElement>(null);
+    const [tooltipPortalTarget, setTooltipPortalTarget] = useState<HTMLDivElement | null>(null);
+
+    useLayoutEffect(() => {
+        setTooltipPortalTarget(tooltipPortalTargetRef.current);
+    }, []);
 
     const [mutationFilterValue, setMutationFilterValue] = useState<MutationFilter>({
         textFilter: '',
@@ -199,7 +204,7 @@ const MutationsOverTimeTabs: FunctionComponent<MutationOverTimeTabsProps> = ({
                             colorScale={colorScale}
                             sequenceType={originalComponentProps.sequenceType}
                             pageSizes={originalComponentProps.pageSizes}
-                            tooltipPortalTarget={tooltipPortalTargetRef.current}
+                            tooltipPortalTarget={tooltipPortalTarget}
                         />
                     ),
                 };


components/src/preact/components/portal-tooltip.tsx
```diff
@@ -0,0 +1,135 @@
...
+ * <PortalTooltip
+ *   content="This is a portal tooltip"
+ *   position="top"
+ *   portalTarget={portalTarget}
+ * >
+ *   <button>Hover me</button>
+ * </PortalTooltip>
+ * ```
+ */
+const PortalTooltip: FunctionComponent<PortalTooltipProps> = ({
```
fengelniederhammer: Nitpick: I prefer to have the "entrypoint" thingy in a file at the top and all helpers and private things below so that I can read the file from top to bottom. Feel free to ignore this if you ordered this file like this on purpose for whatever reason ;)

components/src/preact/components/portal-tooltip.tsx
```diff
@@ -0,0 +1,135 @@
...
+    children,
+    content,
+    position = 'bottom',
+    tooltipStyle,
+    portalTarget,
+}) => {
+    const [isHovered, setIsHovered] = useState(false);
+    const [tooltipPosition, setTooltipPosition] = useState({ top: 0, left: 0 });
```
fengelniederhammer: Why is `tooltipPosition` a state? Shouldn't it be rather a computed property, based on `position` and the refs? Would that be possible at all, since you need to access the refs during render and they might still be `null`?

components/src/preact/components/portal-tooltip.tsx
```diff
@@ -0,0 +1,135 @@
...
+    useLayoutEffect(() => {
+        if (isHovered) {
+            updatePosition();
+        }
+    }, [isHovered, updatePosition]);
+
+    const tooltipContent = (
+        <div
+            ref={tooltipRef}
+            className={`fixed ${TOOLTIP_BASE_STYLES} ${isHovered ? 'visible' : 'invisible'}`}
```
fengelniederhammer: Do we still need to check `isHovered` here? Isn't it already covered by the `isHovered` below (where the portal is created)?

components/src/preact/mutationsOverTime/mutations-over-time.tsx
```diff
@@ -142,6 +142,12 @@ const MutationsOverTimeTabs: FunctionComponent<MutationOverTimeTabsProps> = ({
     overallMutationData,
 }) => {
     const tabsRef = useDispatchFinishedLoadingEvent();
+    const tooltipPortalTargetRef = useRef<HTMLDivElement>(null);
+    const [tooltipPortalTarget, setTooltipPortalTarget] = useState<HTMLDivElement | null>(null);
+
+    useLayoutEffect(() => {
+        setTooltipPortalTarget(tooltipPortalTargetRef.current);
+    }, []);
```
fengelniederhammer: I don't really know whether this is any better, but let's try it :D

fhennig: At least it's not violating the rules! accessing the `.current` is fine in effects, the way I understand it 😁

PR #1021, opened 2025-10-29
feat(components): support multi-select in the lineage filter

fhennig:
### Summary

For https://github.com/GenSpectrum/dashboards/issues/843 we want a lineage filter that allows input for multiple lineages. I wanted to try an implement this here in the components repo.

The `gs-lineage-filter` now has a `multiSelect` boolean switch. Internally, a new `DownshiftComboboxMulti` is used.

What I tried and discarded:
- having a switch between single/multi-select in the components. This seemed to not work very well because to many things were changing
- extracting the multi select functionality from the mutation-filter. This was possible, but then I couldn't easily use it for the lineage filter. They are too different; the mutation-filter also isn't even fetching a list of options; I decided that it would be best to keep that as it's own component.

### Screenshot

https://github.com/user-attachments/assets/fae94b78-2d37-4b9b-9eb0-75f9675f418e

### PR Checklist

### DIFF
--- a/components/src/preact/lineageFilter/lineage-filter.stories.tsx
+++ b/components/src/preact/lineageFilter/lineage-filter.stories.tsx
@@ -61,7 +61,7 @@ const meta: Meta = {
         },
         value: {
             control: {
-                type: 'text',
+                type: 'object',
             },
         },
         width: {
--- a/components/src/web-components/input/gs-lineage-filter.stories.ts
+++ b/components/src/web-components/input/gs-lineage-filter.stories.ts
@@ -79,7 +79,7 @@ const meta: Meta<Required<LineageFilterProps>> = {
         },
         value: {
             control: {
-                type: 'text',
+                type: 'object',
             },
         },
         width: {


fengelniederhammer (commented):
I had a quick look, the general structure looks good, IMO.

Ideally this would reuse some of the things that the mutation filter already has, but I don't know how easily that is possible.

components/src/web-components/input/gs-lineage-filter.tsx
```diff
@@ -125,4 +151,7 @@ type PlaceholderTextMatches = Expect<
     Equals<typeof LineageFilterComponent.prototype.placeholderText, LineageFilterProps['placeholderText']>
 >;
 type WidthMatches = Expect<Equals<typeof LineageFilterComponent.prototype.width, LineageFilterProps['width']>>;
+type MultiSelectMatches = Expect<
+    Equals<typeof LineageFilterComponent.prototype.multiSelect, LineageFilterProps['multiSelect']>
+>;
```
fengelniederhammer: Since you added the types test, we could actually get rid of these.

components/src/web-components/input/gs-lineage-filter.tsx
fengelniederhammer: I'm missing a story that shows the multi mode in the web components storybook.

components/src/preact/lineageFilter/lineage-filter.stories.tsx
```diff
@@ -243,6 +283,145 @@ export const EnterAndClearMultipleTimes: StoryObj<LineageFilterProps> = {
...
+        const canvas = within(canvasElement);
+        const lineageChangedListenerMock = fn();
+
+        await step('Setup event listener mock', () => {
+            canvasElement.addEventListener(gsEventNames.lineageFilterMultiChanged, lineageChangedListenerMock);
+        });
+
+        await step('multi-select filter is rendered with initial values', async () => {
+            await waitFor(async () => {
+                await expect(canvas.getByText('A.1')).toBeInTheDocument();
```
fengelniederhammer: Is there a reason why you didn't use `toBeVisible` in this test?
```suggestion
                await expect(canvas.getByText('A.1')).toBeVisible();
```

components/src/web-components/input/gs-lineage-filter.stories.ts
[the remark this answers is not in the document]
fengelniederhammer: (Same in the Preact story)

Issue #1023, opened 2025-10-31
W-ASAP - Variant explorer doesn't allow selecting "BA.3.2*"


fhennig:
It's not in the options, but BA.3.2.1 and BA.3.2.2 are. I think it probably should be there, right?

PR #1024, opened 2025-11-03
fix(components): add missing lineage prefixes to autocomplete list

fhennig:
This is fixing a reported issue in the dashboards repo: https://github.com/GenSpectrum/dashboard-components/issues/1023

### Summary

Some lineages' parent lineages do not actually appear in the data, for example "BA.3.2" has no tagged lineages on cov-spectrum, but "BA.3.2.1" and "BA.3.2.2" exist. Because of this, the lineage tree definition file does not include "BA.3.2", and so you cannot enter it in the autocomplete list.

The lineage tree definition file does include the parent, but not with this alias name.

This PR looks up all these prefixes in alias lists, and if it can find it, it adds it to the final resulting autocomplete list.

### PR Checklist

### DIFF
--- a/components/src/preact/lineageFilter/fetchLineageAutocompleteList.spec.ts
+++ b/components/src/preact/lineageFilter/fetchLineageAutocompleteList.spec.ts
@@ -321,4 +321,76 @@ describe('fetchLineageAutocompleteList', () => {
             },
         ]);
     });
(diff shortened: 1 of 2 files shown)


fengelniederhammer (approved):
LGTM

PR #1027, opened 2025-11-04
fix(components): make annotation list scrollable

### LINKED ISSUE #1011: mutations over time: 'Filter mutations' popup doesn't handle long list of annotations well
https://github.com/user-attachments/assets/b13ed1ec-a16e-4444-bb1a-86087d589dec

Probably should be a scroll list or something

fhennig:
resolves #1011

### Summary

Add an `overflow-scroll`.

Not very happy with the test, maybe we could also just remove it, not sure if we even need a test here. I wanted to check that initially the last annotation isn't visible, then you scroll, and then it is, but I couldn't get it done.

### Screenshot

https://github.com/user-attachments/assets/f9c65db4-a79f-4d7a-87bf-6e948f7852a6

### PR Checklist

### DIFF
--- a/components/src/preact/components/mutations-over-time-mutations-filter.tsx
+++ b/components/src/preact/components/mutations-over-time-mutations-filter.tsx
@@ -105,7 +105,7 @@ const AnnotationCheckboxes: FunctionComponent<MutationsOverTimeMutationsFilterPr
             <div className='divider mt-0.5 mb-0' />
             <div className='text-sm'>
                 <div className='font-bold mb-1'>Filter by annotations</div>
-                <div className='h-72 overflow-scroll'>
+                <div className='max-h-72 overflow-scroll'>
                     {mutationAnnotations.map((annotation, index) => (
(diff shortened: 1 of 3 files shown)


components/src/preact/components/mutations-over-time-mutations-filter.tsx
```diff
@@ -105,34 +105,36 @@ const AnnotationCheckboxes: FunctionComponent<MutationsOverTimeMutationsFilterPr
...
-                                            annotationNameFilter: new Set(newAnnotationFilter),
-                                        };
-                                    });
-                                }}
-                            />
-                            {annotation.name} (<span className='text-red-600'>{annotation.symbol}</span>)
-                        </label>
-                    </li>
-                ))}
+                <div className='h-72 overflow-scroll'>
```
fengelniederhammer:
```suggestion
                <div className='max-h-72 overflow-scroll'>
```

PR #1031, opened 2025-11-06
fix(components): remove incorrect array type on LineageFilter

fhennig:
### Summary

The `@property({ type: Array })` prevents regular strings from getting set as a value. I think we didn't notice this in our tests, because they bypass this mechanism from the HTML property by setting things directly:

```typescript
const Template: StoryObj<Required<LineageFilterProps>> = {
    render: (args) => {
        return html` <gs-app lapis="${LAPIS_URL}">
            <div class="max-w-(--breakpoint-lg)">
                <gs-lineage-filter
                    .lapisField=${args.lapisField}
                    .lapisFilter=${args.lapisFilter}
                    .placeholderText=${args.placeholderText}
                    .hideCounts=${args.hideCounts}
                    .value=${args.value}
                    .width=${args.width}
                    .multiSelect=${args.multiSelect}
                ></gs-lineage-filter>
            </div>
        </gs-app>`;
    },
    args: {
        lapisField: 'pangoLineage',
        lapisFilter: {
            country: 'Germany',
        },
        placeholderText: 'Enter a lineage',
        value: 'B.1.1.7',
        width: '100%',
        hideCounts: false,
        multiSelect: false,
    },
};
```

I have now added two tests where we set the HTML attributes directly, to check that things get parsed correctly. Example:

```typescript
    render: (args) => {
        return html` <gs-app lapis="${LAPIS_URL}">
            <div class="max-w-(--breakpoint-lg)">
                <gs-lineage-filter
                    lapisField="pangoLineage"
                    placeholderText="Enter a lineage"
                    value="B.1.1.7"
                    .multiSelect=${args.multiSelect}
                ></gs-lineage-filter>
            </div>
        </gs-app>`;
    },
```

I have also added an `updated` override, so we parse arrays correctly when needed. The default type of the property is still string.

For the Preact component underneath, no changes are needed.

---

I also checked, whether we could instead have two properties: `value` and `multiValue` so we wouldn't need this `updated` function. But then there's more complicated stuff going on further down the line, if I want to keep the Preact component the same. And I think in principle in Preact, it makes sense to have a single `value` property, because we can have a union type there. So I'd say that the `value` and `multiValue` approach isn't great.

- [x] covered by an appropriate test

### DIFF
--- a/components/src/web-components/input/gs-lineage-filter.stories.ts
+++ b/components/src/web-components/input/gs-lineage-filter.stories.ts
@@ -155,6 +155,54 @@ export const LineageFilter: StoryObj<Required<LineageFilterProps>> = {
     },
 };
 
+export const LineageFilterStringValue: StoryObj<Required<LineageFilterProps>> = {
+    render: (args) => {
+        return html` <gs-app lapis="${LAPIS_URL}">
+            <div class="max-w-(--breakpoint-lg)">
+                <gs-lineage-filter
+                    lapisField="pangoLineage"
+                    placeholderText="Enter a lineage"
+                    value="B.1.1.7"
+                    .multiSelect=${args.multiSelect}
+                ></gs-lineage-filter>
+            </div>
+        </gs-app>`;
+    },
+    args: {
+        multiSelect: false,
+    },
+    play: async ({ canvasElement }) => {
+        const canvas = await withinShadowRoot(canvasElement, 'gs-lineage-filter');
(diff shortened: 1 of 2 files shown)


chaoran-chen (approved):
I don't really understand the details here and can't judge whether the tests make sense but I'm approving so that the bug can be fixed. @fengelniederhammer, maybe you'd like to take a look once you're back?

PR #1033, opened 2025-11-11
fix(components): use converter instead of updated function

fhennig:
Follow up for https://github.com/GenSpectrum/dashboard-components/pull/1031#issuecomment-3501103308

### Summary

We want to use a converter because this converts the value right at the attribute-property boundary, which is cleaner than using the `updated` function.

### Screenshot
n/a

### PR Checklist

### DIFF
--- a/components/src/web-components/input/gs-lineage-filter.tsx
+++ b/components/src/web-components/input/gs-lineage-filter.tsx
@@ -51,7 +51,7 @@ export class LineageFilterComponent extends PreactLitAdapter {
      */
     @property({
         converter: (value) => {
-            if (value === null || Array.isArray(value)) {
+            if (value === null) {
                 return value;
             }
             try {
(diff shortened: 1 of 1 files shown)


components/src/web-components/input/gs-lineage-filter.tsx
```diff
@@ -49,7 +49,26 @@ export class LineageFilterComponent extends PreactLitAdapter {
...
+                const parsed = JSON.parse(value) as unknown;
+                if (Array.isArray(parsed) && parsed.every((v) => typeof v === 'string')) {
+                    return parsed;
+                }
+                return value;
+            } catch {
+                // If JSON parsing fails, try comma-separated values
+                if (typeof value === 'string' && value.includes(',')) {
+                    return value.split(',').map((s) => s.trim());
+                }
```
fengelniederhammer: Do we really need this? I think we should use less special logic and simply access arrays or strings, nothing else.

fengelniederhammer (approved):

PR #1044, opened 2025-11-21
fix(components): coverage now calculated correctly

### LINKED ISSUE #1041: W-ASAP: Proportion shows 0%, despite 0 Coverage, if samples in timeframe
If there are samples found in the timeframe, but coverage is zero at a given position, the current application shows 0% proportion at a first glance. 

_This gives suggest false security, that the mutation is not there._

We would like to ensure that if no information for that given mutation on a given date exists to show a grey title i.e. N/A. 

Currently, grey tiles are only shown if there is no date in this timeframe. 

The condition should be **timeframe OR Coverage zero**, then show grey.

Image

___

Raised by @AugusteRi, PhD cand. in CBG, thx.

fhennig:
resolves #1041 

### Summary
Add a new `valueWithCoverage` type to `MutationOverTimeMutationValue` that has `count` and `coverage` instead of `proportion`. This means, that even if count is zero, we can still display a coverage value.

We did not have this issue before, because the old method of getting the data never returned count zero and a coverage. It could only return a count, or "below threshold".

I have left the old `value` variant in place for now. I think it's not useful to do a partial refactor of the old code that is due to be removed entirely.

### Screenshot

image

### PR Checklist

### DIFF
--- a/components/src/preact/mutationsOverTime/getFilteredMutationsOverTimeData.ts
+++ b/components/src/preact/mutationsOverTime/getFilteredMutationsOverTimeData.ts
@@ -68,7 +68,7 @@ export function getFilteredMutationOverTimeData({
     if (hideGaps) {
         const dateRangesToFilterOut = filteredData.getSecondAxisKeys().filter((dateRange) => {
             const vals = filteredData.getColumn(dateRange);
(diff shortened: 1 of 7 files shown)


fengelniederhammer (approved):
LGTM 👍

Issue #1057, opened 2025-12-16
Commitlint PR title instead of commits


fengelniederhammer:
and only allow squash merges.

Since we usually squash merge anyway, linting the commit message isn't really relevant. Instead the PR title is relevant.

PR #1060, opened 2025-12-18
chore(github-actions): configure commitlint to lint the PR title

fhennig:
#1057 (not completely resolved: we want to only allow squash-merge)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,6 @@
     "release-please": "^17.1.2"
   },
   "scripts": {
-    "commitlint:last-commit": "commitlint --from=HEAD~1 --verbose",
-    "commitlint:merge-base": "commitlint --from=$(git merge-base HEAD main) --verbose",
     "release-please-dry-run": "release-please release-pr --dry-run --config-file=components/release-please-config.json --manifest-file=components/.release-please-manifest.json --repo-url='https://github.com/GenSpectrum/dashboard-components' --path components",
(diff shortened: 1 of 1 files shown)


fengelniederhammer:
There is also 
```
    "commitlint:last-commit": "commitlint --from=HEAD~1 --verbose",
    "commitlint:merge-base": "commitlint --from=$(git merge-base HEAD main) --verbose",
```
in the top level `package.json` which isn't really necessary anymore. I'm undecided whether to leave it or remove it. What do you think?

fengelniederhammer:
> Maybe the changelog preview could be modified so it does a merge with the PR title to generate the changelog? But there might be merge conflicts. Or maybe there is another way, we could investigate this; but I would just drop this action, personally.

It has been quite useful sometimes (admittedly more when we were new to release-please). Especially for breaking changes it's nice to see how the changelog turns out when you want to write a proper message for breaking changes. But I'd be open to delete it for now.

fengelniederhammer:
> As Fabian said, the preview can be quite useful for people who are not very familiar with the system especially in more complicated cases with breaking changes, but I don't think that it's critical to maintain/adapt it if it takes significant effort.

If something is merged to main that doesn't have the expected changelog, it can still be changed manually in the PR and in the GitHub release. Losing the preview in the worst case means some manual extra work.

fengelniederhammer (approved):

PR #1076, opened 2026-01-29
ci(components): use "trusted publishing" via OIDC for publishing to npm

fengelniederhammer:
### Summary

When creating a new token in npm, it prompted me to use "Trusted publishers" instead: https://docs.npmjs.com/trusted-publishers#supported-cicd-providers I configured the publisher in NPM. Let's see whether it works. Then we wouldn't need the tokens anymore.

### PR Checklist


chaoran-chen (approved):
Sounds good

PR #1083, opened 2026-02-05
feat(components): queries over time label tooltip

### LINKED ISSUE #1074: queries over time: support description text for variant that is shown on hover
As a user of the component I want to describe the queries/collections/co-occurences and have the description visible somewhere. First idea was to have it pop up on hover of the name. We could also have a little asterisk-button next to it (or similar).

cov-spectrum collections already have descriptions that we could then also display.

fhennig:
resolves #1074

### Summary

Adds a new tooltip component that is used on the display labels in the rows. The tooltip displays an optional description, as well as the actual count and coverage queries.

It takes the info from the queries by using the `displayLabel` to look up the full query. This means the `displayLabel`s need to be unique. However, we already did rely on that internally, because the grid component already used the displayLabel as the key in a Map. I've added a `refine` to the schema to enforce display name uniqueness.

### Screenshot

https://github.com/user-attachments/assets/9a76ed23-1e59-408a-b07a-7cee7b3f0e48

### PR Checklist

### DIFF
--- a/components/src/preact/queriesOverTime/queries-over-time.tsx
+++ b/components/src/preact/queriesOverTime/queries-over-time.tsx
@@ -52,13 +52,15 @@ const queriesOverTimeSchema = z.object({
     queries: z
         .array(countCoverageQuerySchema)
         .min(1)
-        .refine(
-            (queries) => {
-                const duplicateDisplayLabels = findDuplicateStrings(queries.map((v) => v.displayLabel));
-                return duplicateDisplayLabels.length === 0;
-            },
-            { message: 'Display labels must be unique' },
-        ),
+        .superRefine((queries, ctx) => {
+            const duplicateDisplayLabels = findDuplicateStrings(queries.map((v) => v.displayLabel));
+            if (duplicateDisplayLabels.length > 0) {
+                ctx.addIssue({
+                    code: z.ZodIssueCode.custom,
+                    message: `Display labels must be unique. Duplicates: ${duplicateDisplayLabels.join(', ')}`,
+                });
+            }
+        }),
     views: z.array(queriesOverTimeViewSchema),
     granularity: temporalGranularitySchema,
     lapisDateField: z.string().min(1),


components/src/preact/queriesOverTime/queries-over-time.tsx
```diff
@@ -136,19 +147,51 @@ const QueriesOverTimeTabs: FunctionComponent<QueriesOverTimeTabsProps> = ({
         });
     }, [queryOverTimeData, proportionInterval, hideGaps, queryFilterValue]);
 
+    const queryLookupMap = useMemo(() => {
+        const map = new Map<
+            string,
+            { displayLabel: string; description?: string; countQuery: string; coverageQuery: string }
```
fengelniederhammer: Isn't that just the inferred type of `countCoverageQuerySchema`? Let's use that instead?

components/src/lapisApi/lapisTypes.ts
```diff
@@ -68,7 +68,7 @@ export const mutationsOverTimeResponse = makeLapisResponse(
 export type MutationsOverTimeResponse = z.infer<typeof mutationsOverTimeResponse>;
 
 const queryDefinition = z.object({
-    displayLabel: z.string(),
+    displayLabel: z.string().optional(),
```
fengelniederhammer: **Critical**: This change creates an inconsistency with the component schema. In `queries-over-time.tsx`, the `countCoverageQuerySchema` requires `displayLabel` to be present (line 43), but here you're making it optional for the LAPIS API.

**Questions:**
1. Does the LAPIS API actually return responses without `displayLabel`? 
2. If yes, how should the component handle missing display labels?
3. If no, this should remain required to match the component's expectations.

**Impact**: If LAPIS can return queries without display labels, the component will fail when trying to use them as Map keys and row labels.

**Recommendation**: Either:
- Keep `displayLabel` required in both schemas if LAPIS always provides it
- Or add proper null-handling throughout the component if it can be missing

components/src/preact/queriesOverTime/queries-over-time.stories.tsx
```diff
@@ -466,6 +466,30 @@ export const WithNoLapisDateFieldField: StoryObj<QueriesOverTimeProps> = {
...
+                countQuery: 'S:346T & S:456L',
+                coverageQuery: '!S:346N & !S:456N',
+            },
+        ],
+    },
+    play: async ({ canvasElement, step }) => {
+        await step('expect error message', async () => {
+            await expectInvalidAttributesErrorMessage(canvasElement, 'Display labels must be unique');
+        });
+    },
```
fengelniederhammer: **FYI**: Excellent test coverage! Adding a story specifically for the duplicate display labels validation is great for ensuring the error handling works correctly. This follows good testing practices.

fengelniederhammer (approved):

PR #1101, opened 2026-03-23
feat(components): gs-mutations-over-time: lazily load pages

### LINKED ISSUE #1093: Mutations over time: lazy loading
As we've considered previously and brought up by @Taepper again yesterday, on a page like https://genspectrum.org/swiss-wastewater/covid, we are loading mutations over time data for thousands of mutations which will not be used by most users. This is rather slow unless the data have been cached. Instead, we should only load the first page by default and load more when needed (when switching pages or downloading everything).

fengelniederhammer:
resolves #1093

### Summary

This adds lazy loading to the `gs-mutations-over-time`. Data fetching is split into two phases:
1. load the overall data, i.e. the columns (dates) and the rows (mutations) of the table
2. fetch the page data based on which rows (mutations) should be shown

Pages are cached, we don't prefetch more that the current page. I don't know yet whether that would be worth the effort.

I also noted layout shifts while loading a page, but decided to split it out into a separate issue.
https://github.com/GenSpectrum/dashboard-components/issues/1102

### PR Checklist


components/src/preact/mutationsOverTime/useMutationsOverTimePageData.ts
```diff
@@ -0,0 +1,88 @@
...
+            lapisFilter,
+            lapis,
+            lapisDateField,
+            sequenceType,
+            requestedDateRanges,
+            pageMutationCodes,
+        );
+        cache.current.set(cacheKey, newData);
+        return newData;
+    }, [lapisFilter, lapis, lapisDateField, sequenceType, requestedDateRanges, pageMutationCodes]);
```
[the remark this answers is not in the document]
fengelniederhammer: I think this should be fine as is? Duplicate request add some server load, but this here should work fine.

components/src/preact/mutationsOverTime/useMutationsOverTimePageData.ts
```diff
@@ -0,0 +1,88 @@
...
+): MutationsOverTimePageQuery {
+    const pageMutationCodes = useMemo(
+        () => filteredMutationCodes.slice(pageIndex * pageSize, (pageIndex + 1) * pageSize),
+        [filteredMutationCodes, pageIndex, pageSize],
+    );
+
+    const cache = useRef<Map<string, MutationOverTimeDataMap>>(new Map());
+    useEffect(() => {
+        cache.current = new Map(); // make sure the cache doesn't grow indefinitely
+    }, [lapisFilter, lapis, lapisDateField, sequenceType, requestedDateRanges]);
```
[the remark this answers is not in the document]
fengelniederhammer: There is a finite amount of pages and without pagination, it also survived it...

components/src/preact/shared/tanstackTable/pagination.tsx
```diff
@@ -13,35 +13,37 @@ export type PageSizes = z.infer<typeof pageSizesSchema>;
...
 
     const minRow = table.getState().pagination.pageIndex * table.getState().pagination.pageSize + 1;
     const maxRow = minRow + table.getRowModel().rows.length - 1;
-    const numRows = table.getCoreRowModel().rows.length;
 
     return (
         <span className='text-sm'>
-            {minRow} - {maxRow} of {numRows}
+            {minRow} - {maxRow} of {totalRows}
         </span>
```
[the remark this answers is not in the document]
fengelniederhammer: Tested it, I didn't find that this is the case

components/src/preact/shared/tanstackTable/pagination.tsx
```diff
@@ -13,35 +13,37 @@ export type PageSizes = z.infer<typeof pageSizesSchema>;
...
 
-function PageIndicator({ table }: PaginationProps) {
-    if (table.getRowModel().rows.length <= 1) {
+function PageIndicator({ table, totalRows }: PaginationProps & { totalRows: number }) {
+    if (table.getRowModel().rows.length <= 1 && totalRows <= 1) {
         return null;
     }
 
     const minRow = table.getState().pagination.pageIndex * table.getState().pagination.pageSize + 1;
     const maxRow = minRow + table.getRowModel().rows.length - 1;
```
fhennig: It's working, but shouldn't this use `getPaginationRowModel`? 

I'd assume that `table.getRowModel().rows.length` gives us the count of all rows (also unfetched ones)

components/src/preact/mutationsOverTime/useMutationsOverTimePageData.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { type MutationOverTimeDataMap } from './MutationOverTimeData';
+import { type MutationsOverTimeMetadata, queryMutationsOverTimePage } from '../../query/queryMutationsOverTime';
+import { type LapisFilter } from '../../types';
+import { Map2dView } from '../../utils/map2d';
+import { useQuery } from '../useQuery';
+
+type MutationsOverTimePageQuery = { isLoading: true; data: null } | { isLoading: false; data: MutationOverTimeDataMap };
+
+export function useMutationsOverTimePageData(
```
fhennig, line 28: Can you write a doc string on this hook? From the whole context I can kind of guess what it's about. This loads data for the mutations over time component, page by page. It caches data as well (what exaclty, when does the cache reset)?

components/src/preact/mutationsOverTime/useMutationsOverTimePageData.ts
```diff
@@ -0,0 +1,88 @@
...
+    return useMemo(() => {
+        if (isLoading) {
+            return { isLoading: true, data: null };
+        }
+
+        return { isLoading: false, data: handleHideGaps(pageData, hideGaps) };
+    }, [pageData, hideGaps, isLoading]);
+}
+
+export function handleHideGaps(data: MutationOverTimeDataMap, hideGaps: boolean) {
```
fhennig, line 96: For completeness, this could also have a little comment on it, something like "returns a new view on the data, where columns are filtered out that do not have any data in them."

fhennig (changes requested):
Looks great!

components/src/preact/shared/tanstackTable/pagination.tsx
fengelniederhammer: Probably doesn't make a difference, but `getPaginationRowModel` seems a bit more intiutive. I changed it.

fhennig (approved):

PR #1113, opened 2026-05-07
fix(components): show aliases as autocomplete options in lineage filter

### LINKED ISSUE #926: Support searching lineages using aliases
It would be nice to allow users to enter aliases when searching for a lineage. In case of SARS-CoV-2, this may mean that it is possible to search for `B.1.1.7.1`.

fhennig:
resolves #926 

## Summary

- Aliases defined in the lineage definition file are now shown as direct autocomplete options in the lineage filter, both as exact matches (e.g. `BA.3.2`) and wildcards (e.g. `BA.3.2*`)
- Aliases are deduplicated case-insensitively, so lowercase noise variants (e.g. `ba.3.2`, `bA.3.2`) are not shown as separate options
- Removes the `findMissingPrefixMappings` function introduced in #1024, which handled a subset of this problem (aliases that happened to be prefixes of other lineage names). The new approach is simpler and covers all alias cases, making the prefix detection logic redundant.

## Background

PR #1024 (d1f34bb) added support for alias-based prefix wildcards (e.g. `BA.3.2*` when `BA.3.2` is an alias of `B.1.1.529.3.2` and `BA.3.2.1` exists in the tree). However, if the alias was not a prefix of any other lineage name, it was silently omitted from the dropdown entirely — which is the case for leaf lineages like `B.1.1.529.3.2` with alias `BA.3.2` when no children use `BA.3.2` as their name prefix.

### DIFF
--- a/components/src/preact/lineageFilter/fetchLineageAutocompleteList.spec.ts
+++ b/components/src/preact/lineageFilter/fetchLineageAutocompleteList.spec.ts
@@ -371,9 +371,31 @@ describe('fetchLineageAutocompleteList', () => {
                 lineage: 'B.1.1.529.3.2*',
                 count: 3,
             },
+            // Long-form aliases of BA.3.2.1 and BA.3.2.2
+            {
+                lineage: 'B.1.1.529.3.2.1',
+                count: 0,
+            },
+            {
+                lineage: 'B.1.1.529.3.2.1*',
+                count: 1,
+            },
+            {
+                lineage: 'B.1.1.529.3.2.2',
+                count: 0,
+            },
+            {
+                lineage: 'B.1.1.529.3.2.2*',
+                count: 2,
+            },
+            // BA.3.2 is the alias of B.1.1.529.3.2, shown as both exact and wildcard
+            {
+                lineage: 'BA.3.2',
+                count: 0,
+            },
             {
                 lineage: 'BA.3.2*',
-                count: 3, // Same as B.1.1.529.3.2* (includes .3.2 and .3.2.1)
+                count: 3,
             },
             {
                 lineage: 'BA.3.2.1',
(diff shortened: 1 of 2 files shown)


components/src/preact/lineageFilter/fetchLineageAutocompleteList.spec.ts
```diff
@@ -393,4 +415,54 @@ describe('fetchLineageAutocompleteList', () => {
...
+            },
+            {
+                lineage: 'B.1.1.529.3.2*',
+                count: 5,
+            },
+            // BA.3.2 is shown as a direct alias option (case-insensitively deduplicated,
+            // so the lowercase variants like ba.3.2, Ba.3.2 etc. are not shown separately)
+            {
+                lineage: 'BA.3.2',
+                count: 0,
```
fengelniederhammer: Why is the count here 0? Shouldn't it also be 5?

fhennig: you are right! I found a bug and fixed it.

components/src/preact/lineageFilter/fetchLineageAutocompleteList.ts
```diff
@@ -43,14 +43,27 @@ export async function fetchLineageAutocompleteList({
...
+            if (seenAliasesUpper.has(aliasUpper)) {
+                return [];
+            }
+            seenAliasesUpper.add(aliasUpper);
+
+            const wildcardCount = getCountsIncludingSublineages(canonicalLineage, lineageTree, countsByLineage);
+            return [
+                { lineage: alias, count: countsByLineage.get(alias) ?? 0 },
+                { lineage: `${alias}*`, count: wildcardCount },
+            ];
```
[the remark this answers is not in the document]
fhennig: let's keep it simple, the trees are not so large.

fengelniederhammer (commented):
Are we sure about this change? As far as I understood it, it adds quite a number of new lineages to the options in the dropdown, doesn't it? In the preview, it also seems a bit laggy, when using all Covid lineages and filtering for `B.` or `B.1`.

fhennig:
Turns out we do not want this.

